PR #27, opened 2022-07-31
change installation from forge

itsanishjain:
## Change the installation from forge

`forge install vectorized/solady`

gives this error

Screenshot from 2022-07-31 08-25-29


Vectorized:
Wait, the forge install is correct. 

You must to be inside a git directory (the `.git` hidden folder must be there), to run the command.

itsanishjain:
opos my bad
I just init the folder with git and it works, thanks for the response

PR #36, opened 2022-08-05
Made it

0xPhaze:
## Description

Gas savings across the board

```
testMulticallablePreservesMsgValueUsedTwice() (gas: -9 (-0.000%)) 
testMulticallablePreservesMsgValue() (gas: -9 (-0.000%)) 
testMulticallableRevertWithMessage(string) (gas: 3 (0.000%)) 
testMulticallableRevertWithMessage() (gas: 3 (0.000%)) 
testMulticallableBenchmark() (gas: -9 (-0.000%)) 
testMulticallableRevertWithCustomError() (gas: 3 (0.000%)) 
testMulticallableRevertWithNothing() (gas: 3 (0.000%)) 
testMulticallableReturnDataIsProperlyEncoded(uint256,uint256,uint256,uint256) (gas: -9 (-0.001%)) 
testMulticallableReturnDataIsProperlyEncoded() (gas: -9 (-0.001%)) 
testMulticallableReturnDataIsProperlyEncoded(string) (gas: -9 (-0.001%)) 
testMulticallablePreservesMsgSender() (gas: -9 (-0.001%)) 
Overall gas change: -51 (-0.003%)
```

Ps: increased optimizoor_runs to 100_000 and included snapshot before ^


0xPhaze:
Actually can go lower when directly encoding in calldata format, though I'm not sure how this would affect internal multicalls..

Included in the latest commit.

```
testMulticallableOriginalBenchmark() (gas: 0 (0.000%)) 
testMulticallableRevertWithNothing() (gas: 24 (0.002%)) 
testMulticallableWithNoData() (gas: -204 (-0.032%)) 
testMulticallableReturnDataIsProperlyEncoded(string) (gas: -756 (-0.059%)) 
testMulticallableReturnDataIsProperlyEncoded(uint256,uint256,uint256,uint256) (gas: -1028 (-0.078%)) 
testMulticallableReturnDataIsProperlyEncoded() (gas: -1028 (-0.079%)) 
testMulticallableBenchmark() (gas: -4357 (-0.123%)) 
Overall gas change: -7349 (-0.370%)
```

foundry.toml
```diff
@@ -7,9 +7,6 @@
 solc_version = '0.8.15'
 auto_detect_solc = false
 optimizer = true
-optimizer_runs = 1_000
+optimizer_runs = 100_000
```
Vectorized, line 10: Left it at 1000 intentionally to test if the code still performs well with smaller bytecode.

src/utils/Multicallable.sol
```diff
@@ -37,22 +42,29 @@ abstract contract Multicallable {
...
-                // Restore `results` and allocate memory for it.
-                results := mload(0x40)
-                mstore(0x40, memPtr)
+
+                return(returndata, sub(memPtr, returndata))
             }
+
+            mstore(0x00, 0x20)
+            mstore(0x20, 0x00)
+            return(0x00, 0x40)
```
Vectorized, line 67: This function is intentionally a public function rather than an external function.

As such, `return` cannot be safely used in assembly.

foundry.toml
0xPhaze: Oke

src/utils/Multicallable.sol
0xPhaze: Yeah makes sense then.

PR #38, opened 2022-08-06
✨ Signature checker

Vectorized:
## Description

Signature checker helper that supports both ECDSA signatures from EOAs and ERC1271 signatures from smart contract wallets.

Note that it's **deliberately** not max optimizoored for gas (something something max-spurious-dragon limit):
- We simply use `ECDSA.recover` instead of inlining it for maintainability.
- We don't hardcode the left padded magic number like `0x1626ba7e00000000000000000...` because of the large increase in bytecode size. So many wasted zeros with just to save a bit of gas with `PUSH32` .
- If you want the max gas efficiency, just use plain old ECDSA.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


src/utils/SignatureCheckerLib.sol
```diff
@@ -0,0 +1,56 @@
...
+                mstore(m, 0x1626ba7e) // `bytes4(keccak256("isValidSignature(bytes32,bytes)"))`.
+                mstore(add(m, 0x20), hash)
+                mstore(add(m, 0x40), 0x40) // The offset of the `signature` in the calldata.
+                mstore(add(m, 0x60), signature.length)
+                calldatacopy(add(m, 0x80), signature.offset, 0x60) // Copy the `signature` over.
+
+                isValid := and(
+                    and(
+                        // Whether the returndata is the magic value `0x1626ba7e` (left-aligned).
+                        eq(mload(add(m, 0x40)), shl(224, mload(m))),
```
Vectorized, line 37: The `mload(m)` is deliberate to avoid a push32 behind the scenes. 

Saves like a 3 gas, but imo not very worth it for the bump in bytecode size.

src/utils/SignatureCheckerLib.sol
```diff
@@ -0,0 +1,56 @@
...
+/// and ERC1271 signatures from smart contract wallets like Argent and Gnosis safe.
+/// @author Solady (https://github.com/vectorized/solady/blob/main/src/utils/Sort.sol)
+/// @author Modified from OpenZeppelin (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/SignatureChecker.sol)
+library SignatureCheckerLib {
+    function isValidSignatureNow(
+        address signer,
+        bytes32 hash,
+        bytes calldata signature
+    ) internal view returns (bool isValid) {
+        address recovered = ECDSA.recover(hash, signature);
```
Vectorized, line 16: Inlining the code may save a little gas, but it is not very worth it for the code duplication, which hurts repo maintainability.

Also, if users truly want max performance, they should just be using plan old ECDSA.

Philosophy: optimize maximally for the best path, but provide good enough performance for other paths.

PR #42, opened 2022-08-10
⚡️ LibRLP optimization and generalization

Vectorized:
## Description

Generalized to numbers from 0 to 2**256-1.

For smaller bytecode and to go light on comments.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


src/utils/LibRLP.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+/// @notice Library for computing contract addresses from their deployer and nonce.
+/// @author Solmate (https://github.com/Rari-Capital/solmate/blob/main/src/utils/LibRLP.sol)
+library LibRLP {
+    // prettier-ignore
+    function computeAddress(address deployer, uint256 nonce) internal pure returns (address deployed) {
+        // @dev The theoretical allowed limit, based on EIP-2681,
+        // for an account nonce is 2**64-2: https://eips.ethereum.org/EIPS/eip-2681.
```
pcaversaccio, line 10: why not a simple check here? The EVM reverts anything above `2**64-2` by definition and instead of a return statement that is technically speaking invalid, we could revert it already here (better to fail loudly than silently). Another idea could be to set `uint64 nonce` and deal with the fact that `type(uint64).max` is still allowed but I guess, that's not something you want to see lol.

Vectorized: Cuz solmate uses uint256 for the nonce, and I don't wanna change it to uint64.

It's for drop-in compatibility.

Vectorized: I can make it return the zero address if it is above the theoretical limit though.

pcaversaccio: why not revert?

Vectorized: After some thoughts, i think it's better to fail silently.

- If the address is to be used after the computation, it is almost certain that the following `create`, `call`, etc, will fail.
- We can't predict if the user is going to pack the address in 32 or 64 bits. While the check may protect against accident dirty upper bits for the 64 bits case, it offers no protection for the 32 bits case. So, it is cleaner to just remove the check and let the user ensure that the nonce is cleaned.

pcaversaccio: The keyword here is _almost certain_. Buy yeah, I get your points. Let's resolve this conversation.

Vectorized: I have added a cautionary comment.

src/utils/LibRLP.sol
```diff
@@ -8,6 +8,8 @@ library LibRLP {
     function computeAddress(address deployer, uint256 nonce) internal pure returns (address deployed) {
         // @dev The theoretical allowed limit, based on EIP-2681,
         // for an account nonce is 2**64-2: https://eips.ethereum.org/EIPS/eip-2681.
+        // Caution! This function will NOT check that the nonce is within the theoretical range.
```
pcaversaccio, line 11: I would write: WARNING

src/utils/LibRLP.sol
```diff
@@ -8,6 +8,8 @@ library LibRLP {
...
+        // It is the user's responsbility to ensure the nonce is valid.
```
pcaversaccio: typo responsibility

src/utils/LibRLP.sol
Vectorized: I left a blank line between the comments to make it clearer.

PR #47, opened 2022-08-11
♻️ Add stronger toString tests

Vectorized:
## Description

Try to corrupt the memory before and after, and see if the string survives.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?

### DIFF
--- a/test/LibString.t.sol
+++ b/test/LibString.t.sol
@@ -20,6 +20,69 @@ contract LibStringTest is Test {
         );
     }
 
+    function testToStringZeroBrutalized() public {
+        string memory s0 = LibString.toString(0);
+        assembly {
+            mstore(mload(0x40), not(0))
+            mstore(0x40, add(mload(0x40), 0x20))
+        }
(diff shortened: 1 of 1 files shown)


test/LibString.t.sol
```diff
@@ -20,6 +20,69 @@ contract LibStringTest is Test {
...
+            mstore(0x40, add(mload(0x40), 0x20))
+        }
+        assertEq(
+            keccak256(bytes(s0)),
+            keccak256(bytes("115792089237316195423570985008687907853269984665640564039457584007913129639935"))
+        );
+        assertEq(
+            keccak256(bytes(s1)),
+            keccak256(bytes("115792089237316195423570985008687907853269984665640564039457584007913129639935"))
+        );
```
hrkrshnn, line 83: So high level solidity can handle the dirty bits and bad encoding well. So the test wouldn't really catch the issues. You'll need to read slots in assembly.

PR #50, opened 2022-08-12
✨ OwnableRoles

Vectorized:
## Motivation

Right now, many code bases mix `Ownable` + `AccessControl`.

However, this requires an extra `SSTORE` on initialization just to give the contract owner the default admin role to grant/revoke roles. This can add a bothersome overhead for clones.

Usually, since the owner will be the only default admin role, the extra `SSTORE` is redundant from a logical point of view.

## Description

This PR combines `Ownable` + `AccessControl` in a minimalist way.

- Roles are represented / set / retrieved as 256 bit words, where each bit represent a single role. This allows us to get and set roles efficiently. Also, it is extremely unlikely that a project will need more than 255 roles. Let me know if you have ever seen a contract with more than 255 roles.

- Compatible with both regular and upgradeable contracts.

- For simplicity, the owner can get and set all roles. If you don't want the owner to get and set certain roles, then simply set the owner to the zero address, and grant the owner a custom admin role instead. Then use the `onlyRoles` modifier on your own custom function to allow the custom admin role to manage only certain roles.

- Follows OZ's `OwnershipTransferred(address indexed oldOwner, address indexed newOwner)` signature. Cuz... yeah, platform compatibility reasons.

- **_Special feature: two step ownership handover._**
  - `function proposeOwnershipHandover(address newOwner) public virtual onlyOwner`
  - `function cancelOwnershipHandover() public virtual onlyOwner`
  - `function acceptOwnershipHandover() public virtual`

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


src/auth/OwnableRoles.sol
```diff
@@ -0,0 +1,599 @@
...
+    /// with both regular and upgradeable contracts.
+    ///
+    /// The role slot of `user` is given by:
+    /// ```
+    ///     mstore(0x00, or(shl(96, user), _OWNER_SLOT_NEG))
+    ///     let roleSlot := keccak256(0x00, 0x20)
+    /// ```
+    /// This automatically ignores the upper bits of the `user` in case
+    /// they are not clean, as well as keep the `keccak256` under 32-bytes.
+    uint256 private constant _OWNER_SLOT_NEG = 0x18207379;
```
Jorropo, line 57: Nit picking but NOT and NEG means two different things.

NOT is a one's complement negation (flip all the bits).
NEG is a two's complement negation (do NOT(x) + 1).

I would call this variable `_OWNER_SLOT_NEG`.

Vectorized: Ok. Sounds good.

src/auth/OwnableRoles.sol
```diff
@@ -0,0 +1,599 @@
...
+    /// with both regular and upgradeable contracts.
+    ///
+    /// The role slot of `user` is given by:
+    /// ```
+    ///     mstore(0x00, or(shl(96, user), _OWNER_SLOT_NOT))
+    ///     let roleSlot := keccak256(0x00, 0x20)
+    /// ```
+    /// This automatically ignores the upper bits of the `user` in case
+    /// they are not clean, as well as keep the `keccak256` under 32-bytes.
+    uint256 private constant _OWNER_SLOT_NOT = 0x18207379;
```
Saw-mon-and-Natalie, line 57: If we had `_OWNER_SLOT_NOT = 0x18207379 << 8 * 28 = 0x1820737900000000000000000000000000000000000000000000000000000000` instead of left-shifting `user` addresses to calculate their `role` storage slot we could have :

```solidity
mstore(0x00, or(user, _OWNER_SLOT_NOT))
let roleSlot := keccak256(0x00, 0x20)
```

without the `newOwner := shr(96, shl(96, newOwner))` bit.

Saw-mon-and-Natalie (changes requested):
@Vectorized , left some comments. Hope it will help 😊.

src/auth/OwnableRoles.sol
```diff
@@ -0,0 +1,599 @@
...
+        }
+    }
+
+    /// @dev Returns whether `user` has all of `roles`.
+    function hasAllRoles(address user, uint256 roles) public view virtual returns (bool result) {
+        assembly {
+            // Compute the role slot.
+            mstore(0x00, or(shl(96, user), _OWNER_SLOT_NOT))
+            // The result is whether the stored value is equal to `roles`.
+            result := eq(sload(keccak256(0x00, 0x20)), roles)
```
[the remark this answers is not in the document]
Vectorized: Wow. Thanks, nice catch! I will change it to the code you suggested.

src/auth/OwnableRoles.sol
Vectorized: I'm quite paranoid about the compiler not cleaning up the upper bits of the addresses. 

Lemme try some test that brutalizes the upper bits of the addresses.

src/auth/OwnableRoles.sol
```diff
@@ -0,0 +1,599 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
```
Vectorized, line 2: I'll change this to `^0.8.4` later.

src/auth/OwnableRoles.sol
```diff
@@ -0,0 +1,599 @@
...
+    }
+
+    /// @dev Receive a two step ownership transfer.
+    /// It will close the handover upon success.
+    function acceptOwnershipHandover() public virtual {
+        assembly {
+            let ownerSlot := not(_OWNER_SLOT_NOT)
+            let handoverSlot := add(ownerSlot, 1)
+            // If the caller is not the handover receipient.
+            if iszero(eq(caller(), sload(handoverSlot))) {
```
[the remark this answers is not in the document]
Vectorized: wow, nice trick!

src/auth/OwnableRoles.sol
Vectorized: From my personal tests, I observe that the compiler can optimize the shifts away if it detects that the addresses are from a clean source (e.g. `msg.sender`), or already clean.

src/auth/OwnableRoles.sol
```diff
@@ -0,0 +1,599 @@
...
+    /// This function must be called upon initialization,
+    /// regardless of whether the contract is upgradeable or not.
+    /// This is to enable generalization to both regular and upgradeable contracts,
+    /// and to save gas in case the initial owner is not the caller.
+    /// For performance reasons, this function will not check if there
+    /// is an existing owner.
+    function _initializeOwner(address newOwner) internal virtual {
+        assembly {
+            // Clean the upper 96 bits.
+            newOwner := shr(96, shl(96, newOwner))
```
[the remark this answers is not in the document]
Vectorized: Oh, the choice to use `shr(96, shl(96, newOwner))` is to reduce the size of the bytecode.

src/auth/OwnableRoles.sol
```diff
@@ -0,0 +1,599 @@
...
+
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                         MODIFIERS                          */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Marks a function as only callable by the owner.
+    modifier onlyOwner() virtual {
+        assembly {
+            // If the caller is not the stored owner, revert.
+            if iszero(eq(caller(), sload(not(_OWNER_SLOT_NOT)))) {
```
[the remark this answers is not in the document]
Vectorized: Hmm... it seems like the compiler optimizes out the `iszero` for `iszero(eq(caller(), sload(not(_OWNER_SLOT_NOT))))`. 

So the `xor` also gives the same gas.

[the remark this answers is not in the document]
Vectorized: Hmm... it seems like the compiler optimizes out the `iszero` for `iszero(eq(caller(), sload(not(_OWNER_SLOT_NOT))))`. 

So the `xor` also gives the same gas.

Vectorized:
@Saw-mon-and-Natalie I've strengthed the tests with some brutalizers.

I am paranoid of the dirty bits in the upper 96 bits of address (e.g. when the addresses come from a packed variable), so I feel it is better to add in the opcodes to clean them. Hopefully, the compiler is smart about it and will clean up the extra opcodes if they are not needed.

I think the compiler will evaluate the `not(_OWNER_SLOT_NOT)`, even without IR. When I tried copy pasting a bytes32 value that is not possible to be simplified (e.g. `0x7ced9b4df95b269341d83b1e9e5c9e4c93ed7f9e329a01520ffaf50a10192ac4`), the gas costs are still the same. I'll just let it be `not(_OWNER_SLOT_NOT)`, such that depending on the optimizer runs, the compiler may choose to evaluate it for better runtime costs, or leave it as it is for smaller bytecode size.

src/auth/OwnableRoles.sol
```diff
@@ -0,0 +1,653 @@
...
+        assembly {
+            // Clean the upper 96 bits, but don't shift it back yet.
+            user := shl(96, user)
+            // Compute the role slot.
+            mstore(0x00, or(user, _OWNER_SLOT_NOT))
+            let roleSlot := keccak256(0x00, 0x20)
+            // Load the current value and `or` it with `roles`.
+            let newRoles := or(sload(roleSlot), roles)
+            // Store the new value.
+            sstore(roleSlot, newRoles)
```
hrkrshnn, line 125: The slot is `keccak((owner << 96) | 0x8b78c6d8))`. Nothing wrong with this by itself. However, if you reuse the same storage scheme in a different contract, the storage collision can become nasty.

A hypothetical example is making an identical copy of `OwnableRoles` and a contract that inherits both of them. The way the compiler handles them is to key off a unique number. The slot for the corresponding mapping would be computed in the following way
```
mstore(0, uniqueid)
mstore(0x20, address)
keccak(0, 0x40)
```
This makes it resilient against collisions by having same logic in parent contracts.

To implement this in OwnableRoles, you could do something like
```solidity
contract OwnableRole {
     uint private HASHINGSALT;
     function hash(address _address) returns  returns (bytes32 ret) {
         assembly {
             mstore(0, HASHINGSALT.slot)
             mstore(0x20, _address)
             ret := keccak(0, 0x40)
         }
     }
}
```

The reason this works is that the storage slot for variable `HASHINGSALT` will always be unique. @Vectorized

hrkrshnn: Also, is it really worth it to do this way? What inefficiencies are you finding in a high level mapping? Perhaps it's easier to fix this in solc if you have specific concerns.

src/auth/OwnableRoles.sol
```diff
@@ -0,0 +1,653 @@
...
+            sstore(ownerSlot, newOwner)
+        }
+    }
+
+    /// @dev Grants the roles directly without authorization guard.
+    /// Each bit of `roles` represents the role to turn on.
+    function _grantRoles(address user, uint256 roles) internal virtual {
+        assembly {
+            // Clean the upper 96 bits, but don't shift it back yet.
+            user := shl(96, user)
```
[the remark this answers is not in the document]
Vectorized: Hmm, but somehow if I don’t do this, the tests which brutalize the upper 96 buts will fail.

src/auth/OwnableRoles.sol
Vectorized: Ah, this idea of using a self provided “hash” for a slot is taken from EIP-2535. 

OwnableRoles is strictly intended to be used on its own too.

Vectorized: Ya, it’s for saving the few gas 😂 .

src/auth/OwnableRoles.sol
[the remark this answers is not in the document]
Vectorized: I’ve added a brutalized test for testInitializeOwnerDirect. If the cleaning in `_initializeOwner` is commented out, then the test will fail.

PR #64, opened 2022-08-28
✨Added Dist Function into FixedPointMathLib

Atarpara:
## Description

Added Absolute Delta Function. @Vectorized may be optimize more can you look this?

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


Vectorized:
Not sure if really want to add this, as it can be confused with https://mathworld.wolfram.com/DeltaFunction.html

Some other functions that can be considered https://docs.python.org/3/library/math.html

- gcd
- lcm
- factorial

Atarpara:
Yes. We can give rename by dist. Simply x-y overflowing on the edge case. 

Example:- 

dist (type(int256).min, type(int256).max) simple x-y overflows the int256 limit. 

thanks for giving me an idea about some more function.

PR #84, opened 2022-09-08
⚡️ Merkle MultiProof optimizations

Vectorized:
## Description

Thanks @wminshew 

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


src/utils/MerkleProofLib.sol
```diff
@@ -56,74 +56,55 @@ library MerkleProofLib {
...
                 // For the case where `proof.length + leafs.length == 1`.
                 if iszero(flags.length) {
                     // If `proof.length` is zero, `leafs.length` is 1.
                     if iszero(proof.length) {
-                        isValid := eq(calldataload(leafsOffset), root)
+                        isValid := eq(calldataload(leafs.offset), root)
                         break
                     }
                     // If `leafs.length` is zero, `proof.length` is 1.
                     if iszero(leafs.length) {
```
wminshew, line 68: this `if` stmt is redundant (and as discussed in tg you could get rid of the other `if` stmt w some `xor`/`mul`s to save a few more gas in this case i think -- but at a minimum you don't need this one)

wminshew: how is a 0-length array handled in calldata? what does `.offset` return?

Vectorized:
```solidity
function cdOffset(uint256[] calldata a, uint256[] calldata b)
    public 
    pure 
    returns (uint256 ao, uint256 bo) 
{
    assembly {
        ao := a.offset
        bo := b.offset 
    }
}
```

a: `[1,2,3]`, b: `[2,3,4,5,6]` => ao: `100`, bo: `228`.  

a: `[]`, b: `[]` => ao: `100`, bo: `132`.

wminshew: ok nice so if the array is empty then `calldataload(offset)` returns `0` it looks like which means since you know one of these arrays has length 1 and one has length 0 you can just do
```
isValid := eq(or(calldataload(leafs.offset),calldataload(proof.offset)), root)
break
```
w/o any nested `if` stmts

Vectorized: Wait, I wouldn't be too sure on that stunt:

```solidity
function testCalldataZeroArrayTrick(uint256[] calldata a, uint256[] calldata b, uint256[] calldata c, uint256[] calldata d) public {
    bool hasError;
    assembly {
        if and(iszero(a.length), gt(calldataload(a.offset), 0)) {
            hasError := 1
        }
        if and(iszero(b.length), gt(calldataload(b.offset), 0)) {
            hasError := 1
        }
        if and(iszero(c.length), gt(calldataload(c.offset), 0)) {
            hasError := 1
        }
        if and(iszero(d.length), gt(calldataload(d.offset), 0)) {
            hasError := 1
        }
    }
    assertFalse(hasError);
}
```

This fuzz test fails.

Vectorized: I'll simply use:

```solidity
isValid := eq(
    calldataload(
        xor(leafs.offset, mul(xor(proof.offset, leafs.offset), proof.length))
    ),
    root
)
```

To be on the safe side.

wminshew: good test & good fix

PR #88, opened 2022-09-12
⚠️ Remove 2 step

Vectorized:
## Description

@pcaversaccio highlighted that a protocol can act as if they have revoked their ownership, but in actuality, restore their ownership via `acceptOwnershipHandover`.

So the trick is:
- `proposeOwnershipHandover`
- `renouceOwnership`
- `acceptOwnershipHandover`

I don't consider this a bug tbh, more like a possible feature for tricking people.

Look, I really don't want to make this code normie-safe by default by making every ownership transfer set the `_proposedOwner` slot to 0, which costs an additional 2.2k gas. 

Plus, it is pure Milady energy to see how people can use this neat "propose-revoke-accept" feature.

Most likely will delete it to prevent hoo-ha.

If you somehow **_still_** need this feature, you can DIY by writing your own function which calls the `_setOwner` internal function. And of course, do not reset the `_proposedOwner` to zero ;)

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


pcaversaccio:
A PoC contract can be found [here](https://goerli.etherscan.io/address/0x5b0c60db3897e6c5a10ad6c2e18556f2d56b97c5).

Vectorized:
Published 0.0.20 on NPM

PR #90, opened 2022-09-12
✨ Add 2 step ownership transfer for OwnableRoles (initiated by receiver)

Vectorized:
## Description

To avoid the extra gas overhead of normal ownership transfers/initialization.

- Basically, instead of the owner initiating the handover, the receiver initiates it via `requestOwnershipHandover()`.
- The owner must type in the correct address when using `completeOwnershipHandover(address)`, or else it will revert.
   - The chance of a typo colliding with a request made by a nefarious spammer is _extremely_ slim. 
- This approach also allows us to avoid having to set a `_pendingOwner` slot to zero on every ownership transfer (2.2k gas) to prevent the bad behavior by devs. (See #88)

Now it’s safe for both devs and speculatoors by default.  

You all can have your cake and eat it. 

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


src/auth/OwnableRoles.sol
```diff
@@ -63,6 +84,13 @@ abstract contract OwnableRoles {
     /// they are not clean, as well as keep the `keccak256` under 32-bytes.
     uint256 private constant _OWNER_SLOT_NOT = 0x8b78c6d8;
 
+    /// The ownership handover request slot of `newOwner` is given by
+    /// ```
+    ///     mstore(0x00, or(shl(96, user), _HANDOVER_SLOT_NOT))
+    ///     let handoverSlot := keccak256(0x00, 0x20)
+    /// ```
+    uint256 private constant _HANDOVER_SLOT_NOT = 0xaccfdaaf;
```
jameswenzel, line 94: The `_NOT` suffix is a bit confusing here, because it's not actually supposed to be bit-flipped to store the handover (like `_OWNER_SLOT_NOT` is) correct?

Is `_HANDOVER_SLOT_SALT` or something a better name?

src/auth/OwnableRoles.sol
```diff
@@ -164,6 +192,52 @@ abstract contract OwnableRoles {
...
+            // Emit the {OwnershipHandoverCanceled} event.
+            log2(0, 0, _OWNERSHIP_HANDOVER_CANCELED_EVENT_SIGNATURE, caller())
+        }
+    }
+
+    /// @dev Allows the owner to complete the handover the ownership to `newOwner`.
+    /// If there is no existing handover requested by `newOwner`, reverts.
+    function completeOwnershipHandover(address newOwner) public virtual onlyOwner {
+        assembly {
+            // Compute and set the handover slot to 0.
```
jameswenzel, line 223: Think this comment is a copy-paste from above 🤠

src/auth/OwnableRoles.sol
```diff
@@ -164,6 +192,52 @@ abstract contract OwnableRoles {
...
+            mstore(0x00, or(shl(96, newOwner), _HANDOVER_SLOT_NOT))
```
jameswenzel, line 224: Am I correct in thinking the only possible collision with a normal storage var in solidity be if you were trying to access a mapping or dynamic array at slot number `(newOwner << 96 | _HANDOVER_SLOT_NOT)` (eg, absurdly large)?

Vectorized: yes

src/auth/OwnableRoles.sol
Vectorized: Could be, but not too strongly opinionated on it.

src/auth/OwnableRoles.sol
Vectorized: Yes hehe

src/auth/OwnableRoles.sol
```diff
@@ -164,6 +194,49 @@ abstract contract OwnableRoles {
...
+            mstore(0x00, or(shl(96, caller()), _HANDOVER_SLOT_NOT))
+            sstore(keccak256(0x00, 0x20), 0)
+            // Emit the {OwnershipHandoverCanceled} event.
+            log2(0, 0, _OWNERSHIP_HANDOVER_CANCELED_EVENT_SIGNATURE, caller())
+        }
+    }
+
+    /// @dev Allows the owner to complete the two-step ownership handover to `newOwner`.
+    /// Reverts if there is no existing ownership handover requested by `newOwner`.
+    function completeOwnershipHandover(address newOwner) public virtual onlyOwner {
```
pcaversaccio, line 221: I would argue that from a security perspective it would make sense to implement a deadline after which the Handover cannot be completed. Ofc from a gas saving perspective you will not like it but I still wanted to add this here :)

Vectorized: Actually, we can do it without much gas impact. 

At least we are not using 256 bits to store a bool.

But the hassle though. Argh..

pcaversaccio: 256 bits are a dream loool - am just afraid that someone DoSes the function `requestOwnershipHandover` with many addresses and the `completeOwnershipHandover` function is called on an address that should not anymore accepted... my concern is more related to chainops topics. Maybe you can open an issue on this and track it there first via discussion...

Vectorized: In such a case, we have to thank the person for keeping ETH deflationary.

Vectorized: After much thought, I think we can do without this feature.

ERC721 has indefinite approvals, and people are somehow still getting by. 

Wouldn't want to complicate the API.

wminshew:
might be missing something (not intimately familiar w codebase) but:

1/ you should only have to delete `_proposedOwner` on renunciations, not every xfer_
2/ if 1/ is wrong for some reason, you could verify owner isn't `0x0` in `acceptOwnershipHandover` (gas overhead should be minimal since owner slot is already touched in txn)

Vectorized:
@wminshew Right now, OZ's new Ownable2Step effectively clears the `_proposedOwner` on every transfer. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable2Step.sol

This also helps prevent the following case:
- Person 'A' propose 2-step ownership handover to person 'C'.
- Person 'A' transfers ownership directly to person 'B', but forgets to cancel the 2-step ownership handover to person 'C'.
- If person 'B' forgets to cancel the 2-step ownership handover too,  person 'C' can make themselves the owner.

Technically speaking, if everyone is careful, there won't be any issues. 

But the choice to use 2-step handover assumes that humans are fallible, so may as well make the handover mechanism as foolproof as possible by default.

wminshew:
the issue then being that you don't enforce the 2-step process? it's merely optional?

wminshew:
not suggesting this necessarily but if so then the workaround could be to, instead of directly storing the proposed new owner's address, to store `keccak256(proposingOwner, proposedOwner)` and then only allow acceptance if `keccak(owner, msg.sender)` matches

Vectorized:
Yup, it’s merely optional. I would prefer direct transfers.

src/auth/OwnableRoles.sol
Vectorized: After some marinating, I have decided to add in a fixed expiry of `block.timestamp + 48 * 3600`.

I'm a shapeshifter.

It also saves a little gas.

src/auth/OwnableRoles.sol
```diff
@@ -164,6 +193,55 @@ abstract contract OwnableRoles {
...
+    /// The request will be automatically expire in 48 hours (172800 seconds) by default.
+    function requestOwnershipHandover() public virtual {
+        unchecked {
+            uint256 expires = block.timestamp + ownershipHandoverValidFor();
+            assembly {
+                // Compute and set the handover slot to 1.
+                mstore(0x00, or(shl(96, caller()), _HANDOVER_SLOT_SEED))
+                sstore(keccak256(0x00, 0x20), expires)
+                // Emit the {OwnershipHandoverRequested} event.
+                log2(0, 0, _OWNERSHIP_HANDOVER_REQUESTED_EVENT_SIGNATURE, caller())
```
pcaversaccio, line 206: maybe `log3` including the expiry date?

src/auth/OwnableRoles.sol
```diff
@@ -193,6 +271,21 @@ abstract contract OwnableRoles {
...
+        assembly {
+            // Compute the handover slot.
+            mstore(0x00, or(shl(96, pendingOwner), _HANDOVER_SLOT_SEED))
+            // Load the handover slot.
+            result := sload(keccak256(0x00, 0x20))
+        }
+    }
+
+    /// @dev Returns how long a two-step ownership handover is valid for in seconds.
+    function ownershipHandoverValidFor() public pure virtual returns (uint64) {
```
pcaversaccio, line 285: just a question for discussion - what you think if you could set that value during construction or initialisation?

src/auth/OwnableRoles.sol
Vectorized: I think not useful. Cuz the indexer can query the expiry

src/auth/OwnableRoles.sol
Vectorized: Prefer this overriding method. I don’t want to bother others in construction or initialisation.

Issue #97, opened 2022-09-15
✨ New LibString functions


Vectorized:
I heard that some people are using LibString in their tests, or tokenURIs.

```Solidity 

uint256 internal constant NOT_FOUND = type(uint256).max;

function slice(string memory s, uint256 start) internal pure returns (string memory result)

function slice(string memory s, uint256 start, uint256 end) internal pure returns (string memory result)

function repeat(string memory s, uint256 times) internal pure returns (string memory result)

function indexOf(string memory s, string memory needle) internal pure returns (uint256 result)

function indexOf(string memory s, string memory needle, uint256 from) internal pure returns (uint256 result)

function lastIndexOf(string memory s, string memory needle) internal pure returns (uint256 result)

function lastIndexOf(string memory s, string memory needle, uint256 from) internal pure returns (uint256 result)

function startsWith(string memory s, string memory needle) internal pure returns (bool result)

function endsWith(string memory s, string memory needle) internal pure returns (bool result)
```

Don't need to max optimize for run-time gas. 

Small code footprint and bytecode size is of priority.

Other common string functions in languages like JS, Ruby, Python have been considered, but are not planned to be added due to the following reasons:
- heavy code needed
- lack of on-chain usefulness
- can be implemented efficiently with native Solidity functions and few lines of code, with the proposed functions

PR #103, opened 2022-09-21
✨ More LibString functions

Raiden1411:
## Description

Added 3 new functions for LibString.sol

-> join
-> concatenate 
-> split   

Also created proper tests in LibString.t.sol.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


Vectorized:
Thanks!

Lemme think about it, cuz it’s kinda pushing the code into the chonky side. Have considered these functions while drafting #97 .

Raiden1411:
Thanks for the consideration!

Looking forward to your reply. Could possibly exclude join from this request since I feel that `concatenate` and `split` would be the most useful ones. Also `split `can't be done in plain solidity.

Vectorized:
I'll see what I can do and modify from here. ;)

PR #118, opened 2022-10-04
✨ Add forceSafeTransferETH + trySafeTransferETH (to combat gas griefing)

Vectorized:
Big thanks to @0xalpharush and @devtooligan for suggesting the `SELFDESTRUCT` trick. 

## Description

To combat gas griefing.

First, we have `forceSafeTransferETH`, which uses `SELFDESTRUCT` via a temporary contract to force send the ETH, if the call exceeds a gas stipend.

Second, we have `trySafeTransferETH`, which returns false instead of reverting, if the call exceeds a gas stipend.

Right now, i've defined some constants as suggestions for the gas stipends:

```solidity
/// @dev Suggested gas stipend for contract receiving ETH
/// that disallows any storage writes.
uint256 internal constant _GAS_STIPEND_NO_STORAGE_WRITES = 2300;

/// @dev Suggested gas stipend for contract receiving ETH to perform
/// a few storage reads and writes, but low enough to prevent griefing.
/// Multiply by a small constant (e.g. 2), if needed.
uint256 internal constant _GAS_STIPEND_NO_GRIEF = 100000;
 ```

If you have a better suggestion for the values, lemme know. Thanks!

Maybe I should add the try variants for the ERC20 stuff too 
(usually ERC20 are from trusted sources, with greatly reduces the risk of grieving, but there could be a need for it).

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?

### CODE REVIEW

**Comment by wilsoncusack** [comment]
> so a griefer can't jam the calling contract

can you say more about what you're thinking of? Successful call that doesn't actual transfer the ETH feels like a foot gun

**Comment by Vectorized** [comment]
@wilsoncusack 

This is for push based refunds.

Consider some kind of auction where non-winners are refunded with a push-based mechanism.

A griever can implement some kind of `receive()` that consumes a ton of gas to make it not possible to refund them. 

This can cause issues if the refunds are in a loop, or via some kind of lazy mechanism (e.g. a new bidder refunds the old bidder who is outbidded).

Personally, I'd prefer pull-based refunds, but some project may have business requirements that heavily favors push based refunds.

**Comment by wilsoncusack** [comment]
Ah ok yeah this makes sense. Have thought of Nouns being griefed in this way. Seems best to use this with a WETH fallback?

**Comment by itsdevbear** [comment]
@Vectorized this is clever, though I think the "try" terminology from a dev UX implies it might "revert, throw, raise an exception" etc.

**Comment by itsdevbear** [comment]
maybe safeTransferEthNoRevert? Kinda long tho

File: src/utils/SafeTransferLib.sol [hunk 1]
```diff
@@ -10,18 +10,37 @@ library SafeTransferLib {
...
+    /// @dev The ERC20 `approve` has failed.
     error ApproveFailed();
 
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                         CONSTANTS                          */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Suggested gas stipend for contract receiving ETH
+    /// that disallows any storage writes.
+    uint256 internal constant _GAS_STIPEND_NO_STORAGE_WRITES = 2300;
```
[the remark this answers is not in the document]
Vectorized: It will be inlined.

src/utils/SafeTransferLib.sol
```diff
@@ -34,10 +53,31 @@ library SafeTransferLib {
...
 
+    /// @dev Sends `amount` (in wei) ETH to `to`, with a `gasStipend`.
+    /// The `gasStipend` can be set to a low enough value to prevent
+    /// storage writes or gas griefing.
+    ///
+    /// Simply use `gasleft()` for `gasStipend` if you don't need a gas stipend.
+    ///
+    /// Note: Does NOT revert upon failure.
+    /// Returns whether the transfer of ETH is successful instead.
+    function trySafeTransferETH(
```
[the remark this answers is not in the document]
Vectorized: I took inspiration from the `tryRecover` functions from OZ's ECDSA.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/ECDSA.sol. 

I know it's kinda quirky though.

src/utils/SafeTransferLib.sol
```diff
@@ -34,10 +53,31 @@ library SafeTransferLib {
...
+        uint256 amount,
+        uint256 gasStipend
+    ) internal returns (bool success) {
+        assembly {
+            // Transfer the ETH and check if it succeeded or not.
+            success := call(gasStipend, to, amount, 0, 0, 0, 0)
+        }
+    }
+
     /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
```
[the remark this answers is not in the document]
Vectorized: Manually handcrafted. Kekw.

src/utils/SafeTransferLib.sol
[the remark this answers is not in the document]
itsdevbear: Yup, just wasn't sure if the compiler was smart enough. Love solc.

src/utils/SafeTransferLib.sol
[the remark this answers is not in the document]
itsdevbear: Yeah don't love that, I think the terminology implies that it might revert. 

I'll meld on it.

src/utils/SafeTransferLib.sol
```diff
@@ -34,10 +53,66 @@ library SafeTransferLib {
...
+                mstore8(0x0b, 0x73) // Opcode `PUSH20`.
+                mstore8(0x20, 0xff) // Opcode `SELFDESTRUCT`.
+                // We can directly use `SELFDESTRUCT` in the contract creation.
+                // We don't check and revert upon failure here, just in case
+                // `SELFDESTRUCT`'s behavior is changed some day in the future.
+                // (If that ever happens, we will riot, and port the code to use WETH).
+                pop(create(amount, 0x0b, 0x16))
+            }
+        }
+    }
```
Vectorized, line 89: Hmm...

0xdineshkumarsm:
- using `mstore(0x00,xor(0x730000000000000000000000000000000000000000ff,shl(8,to)))` reduces gas consumed during the transaction but will increase contract size 
 - `iszero` check can be removed by placing `forcesendEther` in else block .Will this reduce gas?

src/utils/SafeTransferLib.sol
```diff
@@ -34,10 +53,66 @@ library SafeTransferLib {
...
+        uint256 amount,
+        uint256 gasStipend
+    ) internal {
+        assembly {
+            // If insufficient balance, revert.
+            if lt(selfbalance(), amount) {
+                // Store the function selector of `ETHTransferFailed()`.
+                mstore(0x00, 0xb12d13eb)
+                // Revert with (offset, size).
+                revert(0x1c, 0x04)
```
kadenzipfel, line 75: shouldn't this be `revert(0x00, 0x04)`?

src/utils/SafeTransferLib.sol
```diff
@@ -10,18 +10,37 @@ library SafeTransferLib {
...
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Suggested gas stipend for contract receiving ETH
+    /// that disallows any storage writes.
+    uint256 internal constant _GAS_STIPEND_NO_STORAGE_WRITES = 2300;
+
+    /// @dev Suggested gas stipend for contract receiving ETH to perform
+    /// a storage read and a storage write, but low enough to prevent griefing.
+    /// Multiply by a small constant (e.g. 2), if needed.
+    uint256 internal constant _GAS_STIPEND_NO_GRIEF = 30000;
```
kadenzipfel, line 36: imo should note that opcode gas costs may change in the future and that implementers should consider this

Vectorized: Maybe I increase to 100k lol.

src/utils/SafeTransferLib.sol
Vectorized: The selector in the mstore is right-aligned.

But when we return, it must be left-aligned.

clabby: `mstore` always stores bytes < 32b in length as left padded values, and the error selector needs to be right padded in the revert data. Reverting with `mem[28:32]` is the correct way to do this here 😄 

Memory:
```
0x00000000000000000000000000000000000000000000000000000000b12d13eb
```
Revert data:
```
0xb12d13eb
```

Vectorized:
@0xdineshkumarsm I think smaller bytecode is nicer in this case, since nearly all of the time, we will be transferring to non-grievers.

src/utils/SafeTransferLib.sol
Vectorized: Thanks, i just realize there's that pending EIP that suggests doubling SSTORE costs.

src/utils/SafeTransferLib.sol
clabby: The implementation looks sound- this is clever 🫡

They're trying to remove `SELFDESTRUCT` again with [EIP 4758](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4758.md) and it has been gaining some traction recently, but assuming it is implemented in a future fork, the replacement `SENDALL` opcode should not break this function (although, the gas refund from `SELFDESTRUCT` will no longer be there).

devtooligan (approved):
lgtm.  implementation of `forceSafeTransferETH` is 💯 

it still seems a wee bit rash to do this, wyt about doing the WETH step before this and using this as a last resort if the WETH xfer fails? 🤔

Vectorized:
@devtooligan WETH doesn't seem to have a fixed de-facto address across all EVM chains. 

Hardcoding all those addresses can lead to bytecode bloat. 

Also, Solady tries to be as chain agnostic as possible.

devtooligan:
> WETH doesn't seem to have a fixed de-facto address across all EVM chains.
fair

cool, let's do this!!!  🚀 💥

PR #137, opened 2022-10-09
⚡️ Optimize Concat

Raiden1411:
## Description

Optimize the concat function to store `b` in to `a` instead of storing `a` and `b` into new `result` string.

This however creates some problems if you pretend to use the initial value of `a` elsewhere. Solution is to perform that operation first and then use `concat`. This is why the order of `Libstring.t.sol:testStringConcat(string,string)` was changed.

Do you think the gas savings is worth this overhead?
## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


Vectorized:
I initially wanted to try this, but the headache of considering the case where the input is reused made me go with the straightforward approach.

PR #146, opened 2022-10-14
✨ Add DynamicBufferLib

Vectorized:
## Description

Only append.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


RogerPodacter:
Looks awesome!

Quick thought: A nice feature of the old API was that you could call `appendSafe` directly on a `bytes` variable versus having to think about a struct which can be a bit of a mouthful (`DynamicBufferLib.DynamicBuffer memory dynamicBuffer`).

The new API is still better because you don't have to worry about allocating, or "safety," but I wonder if there's a way to preserve the good parts of the old API so you could literally just do:

```solidity
using LibBytes for bytes;

function test() public returns (bytes memory output) {
    bytes[] memory chunks = _getChunks();
    unchecked {
        for (uint256 i; i < chunks.length; ++i) {
            output.append(chunks[i]);
        }
    }
}
```

And then in this case do you even need to call it a "Buffer" or say that it's "Dynamic?" From the user standpoint it's just a way to append bytes to bytes! Let the elite assembly people worry about how it works!

Vectorized:
@RogerPodacter The reason I wrapped it in a struct is so that we don’t need to do the alternative:

```solidity
b = append(b, data);
```

If you have used Rust before, this syntax may be familiar to you. The reason is that the bytes may be reallocated, and the new location must be returned.

Taste wise, i’m ok with that syntax in Rust, but it feels werid for Solidity.

RogerPodacter:
So in that example the `append` would not modify `b` in-place but rather return a new value which would then have to be assigned to `b`?

I agree this is bad and that it's important to preserve the `b.append(data)` syntax, so if the struct is necessary then boom!

PR #151, opened 2022-10-17
✨ add scan forward and batch set/unset to bitmap

estarriolvetch:
## Description

First of all, thank you for building this beautiful library! I may have some of the libraries I wrote based on solady in the future once I test it more :)

This PR ports the scan forward and batch set/unset methods to the bitmap.

Since I haven't learned foundry, I only implement the test it with javascript.

https://github.com/estarriolvetch/solidity-bits/blob/solady/test/bitmap2.js

I will port the tests once I figure out how to do it. It will be welcome if anyone wants to help me on porting the tests!

## Checklist

- [ ] Ran `forge snapshot`?
- [v] Ran `npm run lint`?
- [ ] Ran `forge test`?

### DIFF
--- a/src/utils/LibBitmap.sol
+++ b/src/utils/LibBitmap.sol
@@ -6,6 +6,7 @@ import "./LibBit.sol";
 /// @notice Efficient bitmap library for mapping integers to single bit booleans.
 /// @author Solady (https://github.com/vectorized/solady/blob/main/src/utils/LibBitmap.sol)
 /// @author Modified from Solmate (https://github.com/transmissions11/solmate/blob/main/src/utils/LibBitmap.sol)
+/// @author Modified from Solidity-Bits (https://github.com/estarriolvetch/solidity-bits/blob/main/contracts/BitMaps.sol)
 library LibBitmap {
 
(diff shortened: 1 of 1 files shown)


Vectorized:
Help me see if scanForward can use findFirstSet.

estarriolvetch:
@Vectorized  It's using `fls` rather than `clz` now. It should reduce some operations.

To use `ffs` rather than `fls`, it will require changing the indexing within a bucket of the bitmap.

This will make scan forward more efficient, but contracts uses the old bitmap cannot upgrade to the new version because of the change in the storage layout.

Do you want to make this change?

Vectorized:
I'd like to have the change for consistency, using the lower bits to store the lower-indexed bits.

Vectorized:
@estarriolvetch Ok, i think i know what is `scanForward` doing. 

It is actually scanning the buckets from largest index to smallest index.

In this case, we don't need to use `ffs`.

Imo, it should be renamed to `findGreatestSet` or `findLastSet` for clarity.

Vectorized:
I'll port the tests and merge it in probably a few days.

Thanks!

PR #179, opened 2022-10-24
⚡️ Remove rejection sampling from LibPRNG shuffle

Vectorized:
## Description

Saves 40% gas.

Note that there will be a very tiny modulo bias if the length of the array is not a power of 2.
For all practical purposes, it is negligible and will not be a fairness or security concern.

Thanks @transmissions11 for the meme.

FfzJmaKVsAArQVv

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


wschwab:
imho I'd say have two versions, one more gas optimized, one more randomness optimized

I think it's fair for ppl to want a well optimized randomness-centric implementation, esp off of mainnet, but then also for ppl to want to be able to save that gas on main

Vectorized:
@wschwab After much consideration, I think it is ok to have a little bias.

The maximum size of an array we can shuffle is around 100000. 
The limit is due to quadratic memory expansion costs, and the max block gas limit.

The amount of bias is thus approximately `100000 / (2**128-1) = 2.9387358770557188e-34`.

Which is hopefully small enough for even the most mission critical applications.

PR #193, opened 2022-10-29
⚡️ Reuse offsets and lengths in `LibString.directReturn`

Philogy:
## Description

Noticed you could reuse the return offset and length for the zero-padding. Saves 9 gas. Remaining gas snapshot diff is due to the fuzz tests.

Works because `a.offset + 0x20 + a.length == (a.offset - 0x20) + (a.length + 0x40)` whereby `retOffset = a.offset - 0x20` and `retSize = a.length + 0x40`

## Checklist

- [X] Ran `forge snapshot`?
- [X] Ran `npm run lint`?
- [X] Ran `forge test`?


Vectorized:
Holy shit, no wonder i felt something was off. Thanks!

Issue #152, opened 2022-10-18
Feature Request: Add `BokkyPooBahsDateTimeLibrary`


RogerPodacter:
This is a longish-shot because the lib is moderately niche, but just to stir the pot and get some ideas out there, a snippet I use that's not on NPM and hasn't been updated in a while and is probably unoptimized is [BokkyPooBahsDateTimeLibrary](https://github.com/bokkypoobah/BokkyPooBahsDateTimeLibrary).

Use-case example: [The CryptoPhunk Auction Flywheel](https://www.phunks.pro/) buys NFTs on-demand. The Flywheel uses its ETH balance to buy, but it can't use the entire balance in one week. Instead there is a weekly allowance.

I used BokkyPooBahsDateTimeLibrary to create [a `getStartOfWeek` function that works out this logic](https://etherscan.io/address/0x86b525ab8c5c9b8852f3a1bc79376335bcd2f962#code#F1#L388).

You could imagine this being used in other throttling-type situations, e.g., preventing someone from making 10 DAO proposals in one sitting.

Actually maybe this is a pretty useful library after all...

Vectorized:
Hmm... the library looks small enough. 

May consider.

Vectorized:
@Atarpara You wanna work on this?

PR #209, opened 2022-11-06
✨ Date Time Library

Atarpara:
## Description

Added Date Time Library #152 

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


Atarpara:
@RogerPodacter please check this lib maybe i missed some functions please create issue on https://github.com/Atarpara/DateTimeLib.

Vectorized:
Wow, looks good! The constant should be marked as internal.

Atarpara:
@Vectorized If I marked internal then need to sload in assembly and gas cost increase by 70-80%.

RogerPodacter:
Awesome!!

My quick thought is that it would be nice to have some additional helpers to make it easier to find the number of days / weeks etc between two timestamps and also to be able to find things like "the start of next week" and "the date of the first day of next year."

Even things like `addHours` in the original, which are quite simple, I believe are helpful to users. And the months and years variants are not simple!

Basically the dream would be if this were a full drop-in replacement for Bokky which then people could build on knowing it was the new meme lib to use.

Also, not sure the style guide here, but as a user I'd prefer the default to be that undefined behavior reverts().

Love this!

Vectorized:
I think we can do reverts for undefined cases.

Atarpara:
@RogerPodacter @Vectorized thanks for your feedback. I can start work on this.

Atarpara:
@Vectorized @RogerPodacter I have some question about `submonths()` and `subyears()` functionality. 
Example :- 
Current date is 06-11-2022  if user want to subtract 1 month or 1 year from this date than how many days should be minus.

Vectorized:
@Atarpara need to follow the algorithm in BokkyPooBahs library faithfully, as it deals with leap years.

RogerPodacter:
> Current date is 06-11-2022 if user want to subtract 1 month or 1 year from this date than how many days should be minus.

This is a good point and maybe the function should have a different name like `sameDayIndexInPreviousMonth` (or something better) but I was thinking it should match the colloquial meaning, and when someone says "exactly a month ago" in that case they mean 05-11-2022.

Likewise with year. When someone says "a year ago to the DAY," they are referring to same month / day index in previous year (and are therefore ignoring leap years)

Vectorized:
Disregard my comment if required. I haven’t accumulated enough Schizo energy to burst through this PR yet. Need to marinate more first.

Vectorized:
@Atarpara @RogerPodacter 

Ok, I’ve looked deeper. And I like the design choices. Let’s not revert if the date is invalid. Dev will be responsible for that. This is for bytecode compactnesses. 

I think we don’t need convenience functions that can be easily represented by simple one liner arithmetic expressions.

Vectorized:
@Atarpara What is the purpose of `getNthDayOfWeekInMonthOfYear`?

Atarpara:
@Vectorized yes. For example some dev use `timestampFromDate()` via `isValidDate()` function so max timetamp possible is `type(uint256).max` so for that I consider year max limit.

PR #217, opened 2022-11-12
⚡ Receive sstore 0 gas refund before reverting

MathisGD:
## Description

Weird optimization: when you are going to revert, storing 0 in warm slots costs 100 or 2900 gas but gives a 4800 gas refund, if the value was not already 0. This cannot be the case for `owner`, because the ownership cannot be transferred to address(0).
I would understand if you don't want this kind of optimizations 😄

cc @Vectorized 

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


Vectorized:
wow, lemme think about it. hehe

MathisGD:
Yes that's true. 

The only case where it is important, is when the contracts is expected to revert often. I notably think about liquidation or arbitrage smart-contracts, that reverts when the execution conditions are the ones that were expected.

Vectorized:
After some marinating, I think with the consideration that small bytecode size is one of the goals of solady, such that complex contracts can still fit under the contract size limit, it is better to omit this gas refund trick that isn’t applicable for most use cases. 

Many thanks for this PR.

I’ll credit you if solady ever has functionality in the future that greatly benefits from this trick.

MathisGD:
Cool, no pb ! It is great that you follow precise guidelines to maintain this repo, and know what to merge or not.

PR #219, opened 2022-11-19
🔒 Pass the whole signature length for the EIP1271 signature verification

vladbochok:
## Description

The EIP1271 signature standard provides an abstract interface for smart contracts signature verification.
According to the EIP description, it is a signature format agnostic:
> isValidSignature can call arbitrary methods to validate a given signature, which could be context dependent (e.g. time based or state based), EOA dependent (e.g. signers authorization level within smart wallet), signature scheme Dependent (e.g. ECDSA, multisig, BLS), etc.

It means, that the correct implementation of EIP1271 may accept signatures with arbitrary lengths, not only 65 bytes long. The [Authereum smart contracts](https://github.com/authereum/contracts/blob/master/contracts/account/ERC1271Account.sol#L39) could be an example of EIP1271 implementation that works with `signature.length != 65`. 

Change the `SignatureCheckerLib` implementation to handle arbitrary length signatures.

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


Vectorized:
Big thanks!

PR #220, opened 2022-11-19
🔒 Enable Signature checker to accept arbitary lengths for EIP1271

Vectorized:
## Description

Thanks @vladbochok.

#219

## Checklist

- [x] Ran `forge snapshot`?
- [x] Ran `npm run lint`?
- [x] Ran `forge test`?


src/utils/SignatureCheckerLib.sol
```diff
@@ -34,8 +34,9 @@ library SignatureCheckerLib {
             mstore(m, f)
             mstore(add(m, 0x04), hash)
             mstore(add(m, 0x24), 0x40) // The offset of the `signature` in the calldata.
-            // Copy the `signature` and its length over.
-            calldatacopy(add(m, 0x44), sub(signature.offset, 0x20), 0x61)
+            mstore(add(m, 0x44), signature.length) // The signature length
```
Vectorized, line 37: `signature.length` is already on the stack.

Solidity does some prep for calldata variables -- each calldata array is actually 2 variables on the stack.

src/utils/SignatureCheckerLib.sol
```diff
@@ -34,8 +34,9 @@ library SignatureCheckerLib {
             mstore(m, f)
```
Vectorized, line 34: `f := shl(224, 0x1626ba7e)` will be evaluated at compile time if the number of optimization runs is high enough.

If the optimization runs is low (e.g. 200), even `0x1626ba7e00000000000000000000000000000000000000000000000000000000` will be converted into `0x0b135d3f << 0xe1` by the compiler to minimize bytecode size over runtime gas.

Atarpara: interesting thing.

PR #233, opened 2022-12-03
✨ Add DATA_OFFSET for compatibility with solmate

estarriolvetch:
## Description

Describe the changes made in your pull request here.

Add the DATA_OFFSET constant to the library so it can be a drop-in replacement of solmate's SSTORE2

## Checklist

- [ ] Ran `forge fmt`?
- [ ] Ran `forge snapshot`?
- [ ] Ran `forge test`?


Vectorized:
This is intentional. The underlying data offset should be transparent to the user.

Vectorized:
Nvm, i think we can add it, as it's a library and not an abstract class (no need to be concerned about namespace pollution).

Thanks!

PR #301, opened 2023-01-31
✨ Deterministic SSTORE2

jameswenzel:
## Description

Not sure how best you'd like to integrate this – feel free to make changes as you see fit (or request them). 

It may make more sense to integrate the new methods into the original SSTORE2, but IMO, SSTORE3 is better at implying what this lib is capable of compared to the original 🤠

Notes
- There is duplicated logic (+ bytecode) between `writeDeterministic` and `deriveDeterministicStorageAddress`, used to generate the initCode – this could be broken out into its own internal function, but I believe that would add slight gas overhead due to JUMPs. Thus we are confronted with the age-old dilemma of "gas vs bytecode" 
- The "read" logic is copied from SSTORE2 since it might not make sense to use `SSTORE3` for `writeDeterministic` and then `SSTORE2` to `read`; this is an argument for combining the `writeDeterministic` methods into `SSTORE2`
- when deriving the CREATE2 address, I opted to clobber free memory pointer + zero pointer and restore, rather than expand memory, but you may find this offensive, dangerous, or too presumptive; I considered overwriting first few words of the initCode, but if the data being stored is shorter than 53 bytes (or whatever, if Solidity rounds up the free memory pointer to a multiple of 32, does it?), it would result in more memory being allocated, since 85 bytes are needed to derive the `CREATE2` address.
  - I would not be surprised if you can find some tweaks to optimize it, as well
- Tests are mostly copies of the ones from SSTORE2 – some are redundant, and some may deserve more rigorous testing
## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
I think this can be considered under the umbrella of SSTORE2.

SSTORE2 isn't a real opcode, so let's simply expand it's definition: 
it now allows deterministic deployments optionally with a salt.

CREATE3 abuses a throwaway proxy contract to deploy deterministically, which isn't exactly what is happening here. 
But I get your point.

I think we should include:

```solidity
/// @dev Returns the initialization code hash of the clone of `implementation`.
/// Used for mining vanity addresses with create2crunch.
function initCodeHash(bytes memory data) internal pure returns (bytes32 hash);

/// @dev Returns the address of the deterministic clone of `implementation`,
/// with `salt` by `deployer`.
function predictDeterministicAddress(bytes memory data, bytes32 salt, address deployer)
    internal
    pure
    returns (address predicted)

/*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
/*                      OTHER OPERATIONS                      */
/*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/

/// @dev Returns the address when a contract with initialization code hash,
/// `hash`, is deployed with `salt`, by `deployer`.
function predictDeterministicAddress(bytes32 hash, bytes32 salt, address deployer)
    internal
    pure
    returns (address predicted);

/// @dev Reverts if `salt` does not start with either the zero address or the caller.
function checkStartsWithCaller(bytes32 salt) internal view ;
```

The last 2 functions can be copypastaed from LibClone. 
It's ok if there is a jump, cuz these prediction functions are intended to be called in view functions.

PR #349, opened 2023-03-23
✨ Create ERC1967 Proxy Factory

Vectorized:
## Description

**Outline:**

This is a factory that creates upgradeable ERC1967 proxies. 

Each proxy has one admin. Conceptually, a `mapping(address proxy => address admin)` is stored on the factory (implemented via a functionally similar custom mapping in assembly).

To upgrade a proxy, the admin has to call an upgrade method on the factory, which will call the proxy with the required calldata. The proxy’s bytecode will perform the upgrade if and only if called by the factory.

**Motivation:**

- We can simply use transparent proxy approach instead of UUPS approach. The UUPS approach is intended to eliminate the need for an ProxyAdmin contract, but requires devs to inherit UUPS on their implementation. We simply use the ProxyFactory as the proxy admin contract instead. 
- Almost optimal runtime gas performance (for non admin delegatecalls). The overhead of checking if caller is factory is 21 gas. This is tiny compared to the gas savings from the highly-optimized handcrafted proxy bytecode.
- No need to redeploy a ProxyAdmin contract for every new proxy. 
- No need to verify an admin or proxy contract anymore. Even though the proxy contract is raw bytecode, users can check that it is created by a verified factory and therefore safe.
- Enable easy vanity addresses for upgradeable proxies via create2crunch.

**Todo:**

- [x] Check if Etherscan supports it seamlessly. 
- [x] Write more thorough tests. 
- [x] Mine a canonical create2 vanity address for the factory. Then make a `ERC1967FactoryConstants.sol` that contains that canonical address as a constant, and the bytecode of the deployment (useful for testing purposes via `vm.etch`).

**Deployments:**

Remining a good address. The previous deployment's code wasn't optimal.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/utils/ERC1967Factory.sol
```diff
@@ -0,0 +1,439 @@
...
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                          STORAGE                           */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    // The admin slot for a `proxy` is given by:
+    // ```
+    //     mstore(0x0c, address())
+    //     mstore(0x00, proxy)
+    //     let adminSlot := keccak256(0x0c, 0x20)
+    // ```
```
refcell, line 70: Probably can remove since this is well defined a few lines down.

src/utils/ERC1967Factory.sol
```diff
@@ -0,0 +1,439 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+/// @notice Factory for deploying and managing ERC1967 proxy contracts.
+/// @author Solady (https://github.com/vectorized/solady/blob/main/src/utils/ERC1967Factory.sol)
+/// @author jtriley-eth (https://github.com/jtriley-eth/minimum-viable-proxy)
```
refcell, line 6:
```suggestion
/// @author jtriley-eth (https://github.com/jtriley-eth)
```

src/utils/ERC1967Factory.sol
```diff
@@ -0,0 +1,439 @@
...
+            if iszero(proxy) {
+                mstore(0x00, _DEPLOYMENT_FAILED_ERROR_SELECTOR)
+                revert(0x1c, 0x04)
+            }
+
+            // Set up the calldata to set the implementation of the proxy.
+            mstore(m, implementation)
+            mstore(add(m, 0x20), _IMPLEMENTATION_SLOT)
+            calldatacopy(add(m, 0x40), data.offset, data.length)
+            // Try setting the implementation the proxy and revert upon failure.
```
refcell, line 238:
```suggestion
            // Try setting the implementation on the proxy and revert upon failure.
```

src/utils/ERC1967FactoryConstants.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+/// @notice The address and bytecode of the canonical ERC1967Factory deployment.
+/// @author Solady (https://github.com/vectorized/solady/blob/main/src/utils/ERC1967FactoryLib.sol)
+/// @author jtriley-eth (https://github.com/jtriley-eth/minimum-viable-proxy)
```
refcell, line 6:
```suggestion
/// @author jtriley-eth (https://github.com/jtriley-eth)
```

refcell (commented):
Nice!

src/utils/ERC1967Factory.sol
Vectorized: Oh dear. Time to remine the six zeroes.

src/utils/ERC1967Factory.sol
Vectorized: Hmm… i think we can just leave it.

Vectorized:
Ok, let's go. 

https://etherscan.io/address/0x0000000000000c5b7100000b04dbcb3daceee29c

https://sepolia.etherscan.io/address/0x0000000000ffe8b47b3e2130213b802212439497

https://goerli.etherscan.io/address/0x0000000000000c5b7100000b04DbcB3daCeeE29c#code

https://optimistic.etherscan.io/address/0x0000000000000c5b7100000b04dbcb3daceee29c

https://arbiscan.io/address/0x0000000000000c5b7100000b04dbcb3daceee29c

https://mumbai.polygonscan.com/address/0x0000000000000c5b7100000b04dbcb3daceee29c

https://polygonscan.com/address/0x0000000000000c5b7100000b04dbcb3daceee29c

PR #374, opened 2023-04-11
♻️ ERC209 add view modifier to nonces function

SevenSwen:
## Description
🐞 

I am unable to use the `nonces` function properly in my JS tests because it is lacking the `view` modifier.

- [x] fix nonces function.


Atarpara:
> ## Description
> lady_beetle
> 
> I am unable to use the `nonces` function properly in my JS tests because it is lacking the `view` modifier.
> 
> * [x]  fix nonces function.

It happened due to this function marked virtual. For use this function you need to override this in mock or implement contract. By the way this function must be view. so i think it valid fix. @Vectorized   what is your thought on this?

Vectorized:
Big thanks!!!

PR #380, opened 2023-04-12
Optimization/cashed domain separator

SevenSwen:
## Description
♻️
Before cashed DOMAIN_SEPARATOR:
|	Contract	        | Transfer 	| Tr-rFrom	| Permit    	|
| OpenZeppelin:	|	46620	|	49584	|	74702	|
| SoladyERC20:	|	46370	|	47188	|     ? 76671 ?	|
| SolmateERC20:	|	46505	|	47576	|	74209	| 

After:
| OpenZeppelin:	|	46620	|	49584	|	74702	|
| SoladyERC20:	|	46370	|	47188	|    > 73735 <	|
| SolmateERC20:	|	46505	|	47576	|	74209	|

### DIFF
--- a/src/tokens/ERC20.sol
+++ b/src/tokens/ERC20.sol
@@ -49,12 +49,6 @@ abstract contract ERC20 {
     uint256 private constant _APPROVAL_EVENT_SIGNATURE =
         0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925;
 
-    /// @dev used to recalculate DOMAIN_SEPARATOR in case of network fork
-    uint256 private immutable INITIAL_CHAIN_ID;
-
-    /// @dev default DOMAIN_SEPARATOR
-    bytes32 internal immutable INITIAL_DOMAIN_SEPARATOR;
-
     /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
     /*                          STORAGE                           */
     /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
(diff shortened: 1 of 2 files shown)


SevenSwen:
I understand that this change is not the most elegant, but for now, I haven't come up with anything better.

Vectorized:
Can you do the gas numbers if `name()` is overriden to directly return the string instead of loading it from storage?

SevenSwen:
Do you want me to compare:
```
string private _name;
function name() public view virtual override returns (string memory) {
    return _name;
}
```
and
```
function name() public view virtual override returns (string memory) {
    return "SomeShortString";
}
```
Am I right?

Vectorized:
Yes.

BTW, in some contracts, `name()` may change over time.

SevenSwen:
> in some contracts, name() may change over time.

I agree, but I think it's a trade-off. I believe that 99% of tokens don't want to change name, but 99% of DeFi want to use a cheap permit. In any case, for those tokens that want to change name(), they have the ability to override DOMAIN_SEPARATOR by returning the logic of constant recalculation there.


P.S. I will calculate the gas cost a bit later.

SevenSwen:
`await erc20.estimateGas.name();`
returns:
OpenZeppelin 24279
SolmateERC20 24271
(implementation without using storage) 21566

Vectorized:
Can you calculate the permit costs with the constant name?

SevenSwen:
permit gas costs won't differ due to caching, but the call to computeDomainSeparator will be like this:
with using storage: 25222
without storage: 22502 (21546 if name hardcoded into computeDomainSeparator function)

Vectorized:
I did the tests.

```
| Function Name | min             | avg   | median | max   | # calls |
Solady (storage):
| permit        | 1615            | 30760 | 27276  | 51260 |
Solady (cached):
| permit        | 662             | 29817 | 26335  | 50319 | 
Solady (non-storage):
| permit        | 803             | 29948 | 26464  | 50448 |
Solmate:
| permit        | 709             | 26812 | 26808  | 50820 | 
```

In the case of a non-upgradeable contract (i.e. name will be non-storage), it still offers savings over Solmate.

I think the current code offers a good balance between flexibility and performance.

Vectorized:
Side question:

for this line https://github.com/Vectorized/solady/blob/a59f9bb96ce3d9f9fa3821fad4902a9906afeacb/src/tokens/ERC20.sol#L373

can the `mul(calldatasize(),` be removed? 

If so, why?

 It’s ok to be not sure, cuz even I’m not 100% sure about it right now.

SevenSwen:
It seems like an optimization to avoid an unnecessary MLOAD if 0 is returned.

Issue #396, opened 2023-04-29
✨ Run-length encoding


0xClandestine:
Would love to see more support for data compression, figured I'd draft a RLE lib for [reference](https://en.wikipedia.org/wiki/Run-length_encoding) to get the ball rolling.

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

library RLE {
    function encode(bytes memory data)
        internal
        pure
        returns (bytes memory result)
    {
        unchecked {
            uint256 i = 0;
            uint256 j = 0;
            uint256 encodedLength = 0;

            while (i < data.length) {
                uint256 count = 1;
                while (
                    i + count < data.length && data[i + count] == data[i]
                        && count < 255
                ) {
                    count++;
                }

                encodedLength += 2;
                i += count;
            }

            result = new bytes(encodedLength);
            i = 0;
            j = 0;

            while (i < data.length) {
                uint256 count = 1;
                while (
                    i + count < data.length && data[i + count] == data[i]
                        && count < 255
                ) {
                    count++;
                }

                result[j] = bytes1(uint8(count));
                result[j + 1] = data[i];
                i += count;
                j += 2;
            }
        }
    }

    function decode(bytes memory data)
        internal
        pure
        returns (bytes memory result)
    {
        unchecked {
            uint256 i = 0;
            uint256 j = 0;
            uint256 decodedLength = 0;

            while (i < data.length) {
                uint256 count = uint8(data[i]);
                decodedLength += count;
                i += 2;
            }

            result = new bytes(decodedLength);
            i = 0;
            j = 0;

            while (i < data.length) {
                uint256 count = uint8(data[i]);
                bytes1 value = data[i + 1];

                for (uint256 k = 0; k < count; k++) {
                    result[j] = value;
                    j++;
                }

                i += 2;
            }
        }
    }
}
```

Glad to rewrite this in assembly 🙂

Atarpara:
@0xClandestine have you any good use-case for this?

0xClandestine:
> @0xClandestine have you any good use-case for this?

Certainly, RLE (Run-Length Encoding) is a valuable technique for lossless data compression. I'm currently utilizing it to condense bytecode used for rendering pixel art on-chain. While this use case may seem specialized, I think anything that could reduce storage costs on the EVM is worth exploring. Glad to discuss this further 🙏

RogerPodacter:
[...]

PR #397, opened 2023-05-02
:sparkles: Added RLE encode lib

Atarpara:
## Description

#396 Close.

@Vectorized Need help for testing for code comment.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
I'll clean up and optimize when i have better internet.

Atarpara:
I tried various trick for optimize but unfortunately it will increase gas cost. Glad to see your trick.

0xClandestine:
Just noticed your PR @Atarpara went ahead and pushed some cleanup here: https://github.com/Atarpara/solady/pull/1

Atarpara:
@Vectorized Don't merge this PR. I am thinking to add huffman encoding decoding stuff. Let me know your thought on this.

Vectorized:
@Atarpara I think go with whatever that is:

- robust across various data inputs
- compiles to not too much bytecode
- reasonable runtime gas

Ideally, we want a LibZip.

Atarpara:
@Vectorized  I saw some various lossless compression like huffman coding, entory encoding, deflate I don't  think any of them better than run-length compression in solidity. 
Let me know if  you have any other compression technique in mind and can be better gas perform than RLE.

Vectorized:
I think we should try to implement [this c library](https://github.com/andyherbert/lz1/blob/master/lz.c), and rename this library to LibZip.

- The most common use case for compressing data is to store on-chain NFT pictures. This means compression ratio is more important than runtime gas, cuz the pictures will be uncompressed in a view function.

- Include a `lz77Compress` and `lz77Decompress` which takes in bytes and returns bytes.

- We may include `rleCompress` `rleDecompress` for cases where the data is perfect for RLE compression.

malik672:
how about facebook Zstandard it provides high compression ratio and it's probably faster than lz77

PR #407, opened 2023-05-10
⚡️ Optimize upper bit cleaning with mask

kadenzipfel:
## Description

Implements the use of an address mask to save a couple runtime gas where deployment gas is also reduced. 

This can also be implemented in ERC20.sol and ERC721.sol, but I chose not to include since this method increases deployment size on those contracts. See below gas diffs of ERC20 & ERC721 for decision on inclusion, will add if desired:
- ERC20: https://www.diffchecker.com/cpuEi1Xs/
- ERC721: https://www.diffchecker.com/cpuEi1Xs/

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Atarpara:
Solc optimizer not doing this? Intresting.

Atarpara:
@kadenzipfel Thanks for the PR. `solc` optimizer already doing this. checked `forge snapshot --diff` with main branch there i can't see any gas different adding this changes. So for better readability we keep old version. 

@Vectorized tagging you for double check.

kadenzipfel:
> @kadenzipfel Thanks for the PR. `solc` optimizer already doing this. checked `forge snapshot --diff` with main branch there i can't see any gas different adding this changes. So for better readability we keep old version.
> 
> @Vectorized tagging you for double check.

Hmm, I'm getting different results on `forge test`, should be using the optimizer still, right? Gas savings are pretty minimal and only on applied methods, but here's an example diff with Ownable: https://www.diffchecker.com/R81lABu8/

Vectorized:
Hmm... i prefer the aesthetics of `shr(96, shl(96, a))`.

kadenzipfel:
> Hmm... i prefer the aesthetics of `shr(96, shl(96, a))`.
> 
> Don't need one more constant kekw, which adds to vertical scroll space.

We can always inline it, e.g.:

```
newOwner := and(newOwner, 0x00ffffffffffffffffffffffffffffffffffffffff)
```

but i feel u

PR #429, opened 2023-05-23
`>=` == `!<`

ControlCplusControlV:
## Description

`x >= y` == `x !< y` ∀𝑎∈ℕ

Technically this saves gas in some cases, but not in all. Not sure what you're thoughts are on this but figured I'd make in the PR in case you did want it. This saves gas usually because the compiler only must do `iszero(gt(x,y))` rather than `or(gt(x,y), eq(x,y))`, but can very as its a pure function which will be inlined and optimized

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Atarpara:
@ControlCplusControlV Thanks for PR but solc optimizer already optimize this for us so for better readability we don't merge this.

PR #467, opened 2023-06-07
Use custom error in ERC2981.sol

RogerPodacter:
Amazing library! So nice to not have to worry about the "layout" garbage for storage. There must be a way to make your storage approach used by more people!

I'd be shocked if there wasn't a reason for the require + string thing, but just in case there isn't, I thought I'd get this in there. Note: I do not know assembly so DYOR.

Also I got nervous for upgradeable contracts when I saw the constructor() but looking now I think it's good / harmless.

Honestly if it's any more performant you could consider just hardcoding 10000 as a constant. What kind of sick freak uses a number that's not 10000?

### DIFF
--- a/src/tokens/ERC2981.sol
+++ b/src/tokens/ERC2981.sol
@@ -14,6 +14,9 @@ abstract contract ERC2981 {
 
     /// @dev The royalty receiver cannot be the zero address.
     error RoyaltyReceiverIsZeroAddress();
+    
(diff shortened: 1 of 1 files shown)


Vectorized:
@RogerPodacter The non custom error is intentional for readability, cuz the compiler should optimize it away if the value isn't zero. It's not meant to be used in runtime.

RogerPodacter:
Makes sense! I should have known but I also just learned something! Thank you!

PR #488, opened 2023-06-25
🥢 Fix Ownable comments

mswat5:
## Description
While learning solidity, they said the good way to write function parameters is to start them with an underscore(_). (missed updating the function names, the better way for them will be without (_)).
thanks!
Correct me if I'm wrong.

 optimization      ⚡️

### DIFF
--- a/src/auth/Ownable.sol
+++ b/src/auth/Ownable.sol
@@ -3,9 +3,9 @@ pragma solidity ^0.8.4;
 
 /// @notice Simple single owner authorization mixin.
 /// @author Solady (https://github.com/vectorized/solady/blob/main/src/auth/Ownable.sol)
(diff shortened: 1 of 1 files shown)


Vectorized:
Thanks for finding the typo.

However, on underscores, here is my very strong opinion:

Using underscores for function args is outdated and bad advice. 

Underscores may be used to prefix/postfix local variables to avoid name collision as a last resort. 

Otherwise, it is better is not to prefix/postfix local variables. 
It also helps keep the Natspec on block explorers like Etherscan clean.

Underscores are better reserved for internal/private variables.

Screwing up visibility can mean rekt vs safe. As such, the sacred underscore shall be reserved for a higher purpose of differentiating internal/private vs public/external.

PR #547, opened 2023-08-21
♻️ ERC1967Factory fixes and optimizations

Vectorized:
## Description

- The source code of the canonical deployment https://etherscan.io/address/0x0000000000006396FF2a80c067f99B3d2Ab4Df24#code must and will still be the same due to how `CREATE2` and Etherscan verification works.
- The `ERC1967FactoryConstants.sol` file is updated to contain comments for verification even if we update the source code of `ERC1967Factory.sol`.
- Optimized the custom storage layout of the `proxy => admin` mapping. Use `shl(96, proxy)` in case the public functions are used internally by someone who wants to inherit and make their own factory.
- Added missing 0's into the giant bytecode table comment.
- Make `_initCode()` future safe. Returns `bytes32` instead of `bytes memory`.
- Remove all the `memory-safe` tags. We don't expect this factory to be inherited often anyway.
- Change all instances of `0x89` to `0x88` to remove a redundant stop opcode at the end of the initcode.

Note:

- `predicted` is intentionally left uncleaned to save gas. We have added a note (in the function body, as the function is public) to remind users to clean it as needed.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/utils/ERC1967Factory.sol
```diff
@@ -218,13 +202,12 @@ contract ERC1967Factory {
...
-        bytes memory m = _initCode();
-        /// @solidity memory-safe-assembly
+        bytes32 m = _initCode();
         assembly {
             // Create the proxy.
             switch useSalt
-            case 0 { proxy := create(0, add(m, 0x13), 0x89) }
-            default { proxy := create2(0, add(m, 0x13), 0x89, salt) }
+            case 0 { proxy := create(0, add(m, 0x13), 0x88) }
+            default { proxy := create2(0, add(m, 0x13), 0x88, salt) }
```
Saw-mon-and-Natalie, line 210: note:
- https://github.com/cantinasec/review-solady/pull/4#discussion_r1294072356

for factories with `6` leading zeros.

src/utils/ERC1967Factory.sol
```diff
@@ -276,16 +258,14 @@ contract ERC1967Factory {
     /// @dev Returns the initialization code hash of the proxy.
     /// Used for mining vanity addresses with create2crunch.
     function initCodeHash() public view returns (bytes32 result) {
-        bytes memory m = _initCode();
-        /// @solidity memory-safe-assembly
+        bytes32 m = _initCode();
         assembly {
-            result := keccak256(add(m, 0x13), 0x89)
+            result := keccak256(add(m, 0x13), 0x88)
```
Saw-mon-and-Natalie, line 263: same comment as:
- https://github.com/Vectorized/solady/pull/547/files#r1304847749

src/utils/ERC1967Factory.sol
```diff
@@ -62,12 +62,7 @@ contract ERC1967Factory {
     /*                          STORAGE                           */
     /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
 
-    // The admin slot for a `proxy` is given by:
-    // ```
-    //     mstore(0x0c, address())
-    //     mstore(0x00, proxy)
-    //     let adminSlot := keccak256(0x0c, 0x20)
-    // ```
+    // The admin slot for a `proxy` is `shl(96, proxy)`.
```
Saw-mon-and-Natalie, line 65: why are we shifting?

Vectorized: This is so that in the rare case of someone inheriting the contract to make their own factory, any dirty upper bits in `proxy` will be cleaned out if they call any of the public functions internally.

src/utils/ERC1967Factory.sol
```diff
@@ -250,7 +233,7 @@ contract ERC1967Factory {
             // Store the admin for the proxy.
             mstore(0x0c, address())
             mstore(0x00, proxy)
```
Saw-mon-and-Natalie, line 235: These lines are not needed anymore:
```suggestion
```

PR #548, opened 2023-08-21
♻️ LibClone fixes and optimizations

Vectorized:
## Description

See also: https://github.com/Vectorized/solady/pull/563

- Add a branchless out-of-gas revert if `dataLength > 0xfffd`, which is super unlikely to happen. In practice, `data` will usually, if not always, be the result of an `abi.encodePacked`.
- Add Natspec to further emphasize the `fallback()` difference between this CWIA implementation with the original.
- Place the restoration of the free memory pointer after the `if ..  { revert(..) }`.

Note:
- `predicted` is intentionally left uncleaned to save gas. We have added a note to remind users to clean it as needed.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


WyseNynja:
somewhat related: https://github.com/Vectorized/solady/issues/552

src/utils/LibClone.sol
```diff
@@ -407,8 +410,10 @@ library LibClone {
                 sub(data, 0x3a), 0x9e4ac34f21c619cefc926c8bd93b54bf5a39c7ab2127a895af1cc0691d7e3dff
             )
             mstore(
-                sub(data, 0x5a),
-                or(shl(0x78, add(extraLength, 0x62)), 0x6100003d81600a3d39f336602c57343d527f)
+                // Do a out-of-gas revert if `extraLength` is too big. 0xffff - 0x62 + 0x01 = 0xff9e.
+                // The actual EVM limit may be smaller and may change over time.
+                sub(data, add(0x59, lt(extraLength, 0xff9e))),
+                or(shl(0x78, add(extraLength, 0x62)), 0xfd6100003d81600a3d39f336602c57343d527f)
```
Vectorized, line 416: The initcode has been prefixed with a `0xfd` (revert), which will cause an EVM error (consumes all gas), as there are insufficient stack variables.

If `lt(extraLength, 0xff9e)` is true, then the start of the initcode will be `fd`, otherwise it will be `61`.

src/utils/LibClone.sol
```diff
@@ -506,6 +516,10 @@ library LibClone {
             let dataEnd := add(add(data, 0x20), dataLength)
             let mAfter1 := mload(dataEnd)
 
+            // Do a out-of-gas revert if `dataLength` is too big. 0xffff - 0x02 - 0x62 = 0xff9b.
+            // The actual EVM limit may be smaller and may change over time.
+            returndatacopy(returndatasize(), returndatasize(), gt(dataLength, 0xff9b))
```
Vectorized, line 521: Since we aren't deploying any contract, we can't use the same trick in the `clone` and `cloneDeterministic` functions. Hence, we'll use this alternative instead.

Saw-mon-and-Natalie:
LGTM 🎊

PR #555, opened 2023-08-22
⚡️ Optimize ETH transfers

Vectorized:
## Description

Saves 2 gas. 

Not all chains have PUSH0.

Some may ask: why not just use PUSH0 for mainnet and no PUSH0 for L2s?

Many hyper-optimized contracts use CREATE2 to deploy to canonical addresses. Hence the initcode must be the same. Hence, it will be good to have it optimized for EVMs without PUSH0.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/utils/SafeTransferLib.sol
```diff
@@ -91,12 +91,12 @@ library SafeTransferLib {
...
                 revert(0x1c, 0x04)
             }
-            if iszero(call(gasStipend, to, amount, 0x00, 0x00, 0x00, 0x00)) {
+            if iszero(call(gasStipend, to, amount, gas(), 0x00, gas(), 0x00)) {
                 mstore(0x00, to) // Store the address in scratch space.
                 mstore8(0x0b, 0x73) // Opcode `PUSH20`.
                 mstore8(0x20, 0xff) // Opcode `SELFDESTRUCT`.
                 if iszero(create(amount, 0x0b, 0x16)) {
-                    if iszero(gt(gas(), 1000000)) { revert(0x00, 0x00) } // For gas estimation.
+                    returndatacopy(gas(), returndatasize(), shr(20, gas())) // For gas estimation.
```
Vectorized, line 99: The `returndatacopy(gas(), returndatasize(), shr(20, gas()))` causes an out-of-gas revert if the remaining gas is less than 1048576.

This makes the gas estimator try increasing amounts of gas until the `create` operation has enough gas to definitely execute fully. Currently, it is about 5k gas. It is super unlikely that we will have a future where the EVM is changed to make it cost more than 1 million gas.

PR #622, opened 2023-09-21
✨ Receiver

z0r0z:
## Description

Maybe not-so-cursed token receiver callback handling via fallback().

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


atarapra (commented):
This function would never revert this is expected behaviour? This can be lead issues like permanent ETH lock, successfully pass any contract call, etc..

z0r0z:
@atarapra My understanding is that revert behavior for `safe` transfers are managed through the tokens themselves. Correct me if wrong (as I find this pattern interesting), doesn't all this fallback() do is, if someone calls the contract with a receiver selector, it just returns that same selector back? As for eth receipts, I think *as is* this would behave like a receive() but that should probably be included in this contract too. We could add a revert at end though if you think that would benefit.

benaadams:
Shouldn't be `payable` I think is the point that @atarapra is making or the contract now accepts ETH and has no handling for it

Vectorized:
I think this contract is meant to be inherited by smart wallets. Maybe combining it with the smart wallet will be better.

src/utils/Receiver.sol
```diff
@@ -0,0 +1,30 @@
...
+contract Receiver {
+    fallback() external payable {
+        assembly {
+            // Shift right by 224 bits to get the function signature.
+            let shiftedSig := shr(224, calldataload(0))
+
+            // `keccak256("onERC721Received(address,address,uint256,bytes)")`.
+            if eq(shiftedSig, 0x150b7a02) {
+                mstore(0x00, shl(224, shiftedSig))
+                return(0x00, 0x20)
```
Vectorized, line 14:
```solidity
mstore(0x40, shiftedSig) 
revert(0x3c, 0x20)
```

Vectorized, line 14:
```suggestion
                mstore(0x20, shiftedSig)
                return(0x3c, 0x20)
```
The upper 224 bits of 0x40 will usually, if not always, be zero if the function dispatcher hits the fallback function.

z0r0z:
I also tentatively put forth this future-proof pattern that handles those certain tkn 'safety' standards that attempt to insert such usefulness as 'if you send a bytes4, you get it back, so it's safe' ~~ but I think the PR's opinionation to known magic vals might be palatable to start for wallet contractors:

`fallback() external payable {
        assembly {
            mstore(32, shr(224, calldataload(0)))
            return(60, 32)
        }
    }`

Vectorized:
i prefer the future-proof pattern.

Vectorized:
@z0r0z after some thought, i think let's stick to the non-future proof pattern. This is so that users still can use stuff like `LibZip.cdFallback()`.

I strongly think that any token ERCs from now onwards will not use that kind of "safe" pattern, or it will not be compatible with the plethora of wallets out there.

z0r0z:
that makes sense @Vectorized ~~ you sniped me midchange, but now there is a metric to compare against, so that's good. Revert'n.

src/accounts/Receiver.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+/// @notice Universal transfer callback receiver using fallback returns of selector calldata.
```
Vectorized, line 4:
```suggestion
/// @notice Receiver mixin for ETH and safe-transferred ERC721 and ERC1155 tokens.
```

Vectorized:
LGTM.

src/accounts/Receiver.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+/// @notice Receiver mixin for ETH and safe-transferred ERC721 and ERC1155 tokens.
+/// @author Solady (https://github.com/Vectorized/solady/blob/main/src/accounts/Receiver.sol)
+///
+/// @dev Note:
+/// - Handles all ERC721 and ERC1155 token safety callbacks.
```
atarapra, line 8: Add note: this contract has fallback function so any arbitrary call with non empty calldata will successfully execute.

z0r0z:
> I also tentatively put forth this future-proof pattern that handles those certain tkn 'safety' standards that attempt to insert such usefulness as 'if you send a bytes4, you get it back, so it's safe' ~~ but I think the PR's opinionation to known magic vals might be palatable to start for wallet contractors:
> 
> `fallback() external payable { assembly { mstore(32, shr(224, calldataload(0))) return(60, 32) } }`

for posterity and for future consideration if using the above version, `payable` `fallback()` maxxing by placing a `callvalue()` helper in first line instead of having `receive()` in a contract ends up saving more gas for contract deployment (smaller code) and for safetransfer token receipts:

```
fallback() external payable {
        assembly ("memory-safe") {
            if callvalue() { return(0x00, 0x00) } // `receive()`.
            mstore(0x20, shr(224, calldataload(0)))
            return(0x3C, 0x20)
        }
    }
```

PR #623, opened 2023-09-21
✨ EIP4337 Simple Account

z0r0z:
## Description

Starting on base account and other EIP4337 infra.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


atarapra:
@z0r0z if you want help I can help you.

Vectorized:
Ok, after some thought, here are my opinions:

- We won't use UUPS. Our `EIP4337.sol` (akin to infinitism's `SimpleAccount.sol`) is intended to be published as an implementation, and the canonical `ERC1967Factory` will be used to clone, initialize, and upgrade it.

- `_entryPoint()` can be a pure function that simply returns a constant. Currently, the canonical entry point is 0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789.

- I think that the `fallback` function of the `EIP4337.sol` should call `Receiver.sol`'s `fallback` first, followed by `LibZip.cdFallback()`. Most `EIP4337.sol` will be used on L2, which means calldata compression will be useful. 

- Not sure if it should be called `EIP4337` or `ERC4337`.

z0r0z:
This seems simple to start. To follow rest of Solady format `_ERC_4337` seems onpoint since I can't imagine much in way of finalizing (on that note prob should prob rename `_EIP_712.sol` -> `ERC712.sol`). Using canon ERC1967 factory is convenient but I think since contract accounts are the most numerous contract use case a more optimized and opinionated factory (which gets redeployed to a constant implementation?) would be advantageous.

z0r0z:
should we bake ERC712 into this or just `toEthSignedMessageHash()`? see https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/samples/SimpleAccount.sol#L102

Vectorized:
> should we bake ERC712 into this or just `toEthSignedMessageHash()`? see https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/samples/SimpleAccount.sol#L102

Whichever requires lesser code for now.

PR #625, opened 2023-09-27
⚡️  ECDSA potential optimizations

z0r0z:
## Description

Suggestions as snips and nits to save gas() in various ecrecover functions. Security comments useful obv. I'm just snippin` over here.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


atarapra:
@z0r0z  good catch. I will not recommend changes because it is save very tiny amount of gas(around 20-25), it poses a security risk.Let's consider a hypothetical example of a contract that uses Solady's ECDSA library and a storage mapping of signature hashes. (please ignore nonce protection for this contract) 

```
import {ECDSA} from "solady/utils/ECDSA.sol";

contract Signature {
          mapping (bytes32 sigHash => bool used) public signs;
          
          
          function name1(bytes memory signature) public {
          
                   // checking signature is not used
                  require(!signs[keccak256(signature)], "already used");
          
                  // do some stuff
          
                 // assume ECDSA is solady  lib 
                ECDSA.recover(msgHash, signature);   
           
                signs[keccak256(signature)] = true;
          }
}
```

In the above given contract where an attacker can append random data to the end of a valid signature.

@Vectorized what you think about this?

z0r0z:
Thanks for checking this out @atarapra ~~ I def appreciate the subtle way the library distinguishes between EIP- 2098 compact signatures and ordinary 65 byte-length ones with that check (and the return (1) as ecrecover dest is cool), but I think there might not be a problem as far as security goes, thankfully, because if users provide more bytes than expected, those are simply ignored and not passed into the 128 bytes sent into ecrecover staticcall? For sure respect mind towards security, but I guess if Solady doesn't enforce 64 bytes in EIP-2098 case, why enforce in 65 bytes vanilla case?  🚬 👓 💭  Maybe we add a comment putting it on users to ensure extra bytes don't get used by their smart contract in a weird way, but it might be ok overall as this function just recovers and cleans context after.

atarapra:
@z0r0z for EIP2098 signature types there are separate function and it's enforcing 64 bytes in two separate parameters. Putting comments is fine but most off time dev didn't read comments. Currently solady has too much comments for these type of common mistake.

Vectorized:
I'll leave the check in. 

EIP-2098 is implemented via the (r,sv) variant.

z0r0z:
fair, but what do ya think of first nit @Vectorized: https://github.com/Vectorized/solady/pull/625/commits/02e181027fea5f98e30616f8fa31ca79e82a75bd

lowers gas and codesize so looks legit to me

Vectorized:
@atarapra @z0r0z 

https://github.com/Vectorized/solady/blob/main/audits/cantina-solady-report.pdf 

Scroll down to see 4.2.3: ECDSA: Empty signature can result in valid recovered address.

Finding by @Philogy.

Vectorized:
Most likely won't merge, but leaving open for some time.

atarapra:
@Vectorized This PR first commit is valid for saving gas. @z0r0z could revert all other changes and we good to go.

image

PR #629, opened 2023-10-04
✨ Add Pausable mixin, which allows children to implement an emergency stop

eugenioclrc:
## Description

Pausable mixin, which allows children to implement an emergency stop.
This mixin is used through inheritance. It will make available the modifiers `whenNotPaused` and `whenPaused`, which can be applied to the functions of your contract. Note that they will not be pausable by simply including this module, only once the modifiers are put in place.
This implementation is inspired on the OpenZeppelin implementation.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
Looks good. Might merge.

OZ's implementation's pretty efficient tho, so... unless there's something I've missed.

@atarapra comments?

eugenioclrc:
:D thanks for the compliment, means a lot to me!

image

atarapra:
Q@eugenioclrc @Vectorized left comment on pause solt. What you think about that?

src/utils/Pausable.sol
```diff
@@ -0,0 +1,135 @@
...
+            mstore(0x00, caller())
+            log1(0, 0x20, _PAUSABLE_UNPAUSED_EVENT_SIGNATURE)
+        }
+    }
+
+    /// @dev Throws if the contract is paused.
+    function _checkNotPaused() internal view virtual {
+        /// @solidity memory-safe-assembly
+        assembly {
+            // If the caller is not the stored owner, revert.
```
romeroadrian, line 88: Wrong comment?

src/utils/Pausable.sol
```diff
@@ -0,0 +1,135 @@
...
+                revert(0x1c, 0x04)
+            }
+        }
+    }
+
+    /// @dev Throws if the contract is not paused.
+    function _checkPaused() internal view virtual {
+        /// @solidity memory-safe-assembly
+        assembly {
+            // If the caller is not the stored owner, revert.
```
romeroadrian, line 100: idem above, looks like an incorrect comment

src/utils/Pausable.sol
```diff
@@ -0,0 +1,135 @@
...
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Emitted when the pause is triggered by `account`.
+    event Paused(address account);
+
+    /// @dev Emitted when the pause is lifted by `account`.
+    event Unpaused(address account);
+
+    /// @dev `keccak256(bytes("Paused(address)"))`.
+    uint256 private constant _PAUSABLE_PAUSED_EVENT_SIGNATURE =
```
romeroadrian, line 35: Can these constants be the actual expression instead of the resolved value? or solidity won't resolve these at compile time? Should improve clarity and avoid the ugly chunk of numbers

e.g.
```
uint256 private constant _PAUSABLE_PAUSED_EVENT_SIGNATURE = keccak256(bytes("Paused(address)"));
```

src/utils/Pausable.sol
```diff
@@ -0,0 +1,135 @@
...
+        assembly {
+            _paused := eq(sload(not(_PAUSED_SLOT_NOT)), _PAUSED_STATE)
+        }
+    }
+
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                         MODIFIERS                          */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Marks a function as only callable by the owner.
```
romeroadrian, line 124: Wrong comment

src/utils/Pausable.sol
```diff
@@ -0,0 +1,135 @@
...
+    modifier whenPaused() virtual {
+        _checkPaused();
+        _;
+    }
+
+    /// @dev Marks a function as only callable by the owner.
```
romeroadrian, line 130: one more..

romeroadrian (commented):
LGTM

src/utils/Pausable.sol
eugenioclrc: Comment now is:
```
            // If the contract is not in `_PAUSED_STATE`, revert.
```

src/utils/Pausable.sol
eugenioclrc: Must say that i agree with you but the current convention in the others falis is to put the resolved expression.

src/utils/Pausable.sol
eugenioclrc: New comment:
```
    /// @dev Marks a function as only callable if the contract is paused.
```

src/utils/Pausable.sol
eugenioclrc: New comment:
```
    /// @dev Marks a function as only callable if the contract is not paused.
```

src/utils/Pausable.sol
eugenioclrc: Wdyt about this @Vectorized ?

Vectorized:
I just remembered why I didn’t add this. It’s so that people are forced to implement their own packed pausable.

src/utils/Pausable.sol
Vectorized: Put the resolved expression.

atarapra:
@Vectorized I agree with you. @eugenioclrc Thanks for the great PR but unfortunately we will not merge this. OpenZeppelin pausable version is also efficient if someone really need extra slot for this thing.

eugenioclrc:
Hooo so much pain :(, is not the first time im rejected, ill be back.

Thanks take a look into it, and especial thanks to @romeroadrian to review it.

PR #637, opened 2023-10-08
✨ Add UUPS Upgradeable

atarapra:
## Description

Add UUPS Pattern for smart contract upgrade. 
 
## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/utils/UUPSUpgradeable.sol
```diff
@@ -0,0 +1,84 @@
...
+
+            if iszero(
+                eq(
+                    mload(staticcall(gas(), newImplementation, 0x1c, 0x04, 0x01, 0x20)),
+                    ERC1967_IMPLEMENTATION_SLOT
+                )
+            ) {
+                mstore(0x00, 0x55299b49) // error selector of UpgradeFailed()
+                revert(0x1c, 0x04)
+            }
```
Vectorized, line 45:
```suggestion
            mstore(0x01, 0x52d1902d) // bytes4(keccak256("proxiableUUID()"))

            if iszero(
                eq(
                    mload(staticcall(gas(), newImplementation, 0x1d, 0x04, 0x01, 0x20)),
                    ERC1967_IMPLEMENTATION_SLOT
                )
            ) {
                mstore(0x00, 0x55299b49) // `UpgradeFailed()`.
                revert(0x1c, 0x04)
            }
```

src/utils/UUPSUpgradeable.sol
```diff
@@ -0,0 +1,84 @@
...
+/// @author Modified from OpenZeppelin (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/utils/UUPSUpgradeable.sol)
+abstract contract UUPSUpgradeable {
+    /// @dev The upgrade failed.
+    error UpgradeFailed();
+
+    /// @dev The ERC-1967 storage slot for the implementation in the proxy.
+    /// `uint256(keccak256("eip1967.proxy.implementation")) - 1`.
+    bytes32 internal constant ERC1967_IMPLEMENTATION_SLOT =
+        0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
+
```
Vectorized, line 15:
```suggestion
    bytes32 internal constant _ERC1967_IMPLEMENTATION_SLOT =
        0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;

```

src/utils/UUPSUpgradeable.sol
```diff
@@ -0,0 +1,84 @@
...
+                returndatacopy(0x00, 0x00, returndatasize())
+                return(0x00, returndatasize())
+            }
+        }
+    }
+
+    /// @dev Upgrades the implementation of th proxy to `newImplementation`.
+    /// The caller of this function must be `_authorizeUpgrade`.
+    function _upgradeTo(address newImplementation) internal virtual {
```
Vectorized, line 66: Change to public virtual. 

See: https://goerli.etherscan.io/address/0xf9dc117165100cdeae813e49e20d0d5a055e9e86#writeContract

Vectorized (changes requested):
Great work!

Vectorized:
Should we emit the event for upgrade? Not compulsory though. 

https://eips.ethereum.org/EIPS/eip-1967#logic-contract-address

Reason for not emitting: save gas.

Reason for emitting: Most 4337 wallets will be on L2, where gas is cheap. Emitting can help making tracking upgrades easier.

atarapra:
Thinking more about upgraded event, it seems that it's not obligatory according to the (EIP). If someone desires to include the upgrade event, they have the option to override the function. You are right, Layer 2 gas costs are indeed economical in terms of gas prices, Solady continues to be well-regarded for its gas optimization capabilities, regardless of whether it's on Layer 1 or Layer 2. If we can save 1 gas then still I will go for it.

What are your thoughts on this?

atarapra:
Update: Etherscan automatically tracks contract upgrades without emitting a separate upgrade event. Additionally, it provides visibility into the previous implementation of the contract. So with help of etherscan users can track easily one previous implementation.

image

Vectorized (commented):
I think let's add the event.

The ERC1967Factory does emit a event, although not on the actual proxy itself.

Also, let's clear the upper 96 bits of the address.

Vectorized (approved):

Vectorized:
Optimizing for smaller bytecode size since upgrades are not intended to be often on L1s (e.g. in Defi singleton contracts), and are cheap on L2s (e.g. in 4337 wallets).

PR #639, opened 2023-10-11
✨ ERC4337 Simple Account

Vectorized:
## Description

ERC4337 Simple Account.

Design:
- Recycle `Ownable` to double as a initialization guard.
- Not super-inlined, intentional for brevity and readability. This standard is already quite complex.
- Includes a `delegateExecute(address delegate, bytes calldata data)` function for on-the-fly plugins.

Currently omitted functions:

- `getNonce()`
- `getNonce(uint192 key)`
- `incrementNonce(uint192 key)`
- `supportsInterface(bytes4)` (do people actually use this on-chain?)

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,292 @@
...
+    /// it will not be required to send again)
+    ///
+    /// `missingAccountFunds` is the minimum value this method should send the EntryPoint,
+    /// which MAY be zero, in case there is enough deposit, or the userOp has a paymaster.
+    function _payPrefund(uint256 missingAccountFunds) internal virtual {
+        /// @solidity memory-safe-assembly
+        assembly {
+            if missingAccountFunds {
+                // Ignore failure (its EntryPoint's job to verify, not the account's).
+                pop(call(not(0), caller(), missingAccountFunds, codesize(), 0x00, codesize(), 0x00))
```
leekt, line 112: you can use gas() instead of not(0) here actually

```
Must not use GAS opcode (unless followed immediately by one of { CALL, DELEGATECALL, CALLCODE, STATICCALL }.)
```

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,292 @@
...
+    {
+        bool sigFailed = owner()
+            != ECDSA.recoverCalldata(ECDSA.toEthSignedMessageHash(userOpHash), userOp.signature);
+        /// @solidity memory-safe-assembly
+        assembly {
+            // Returns 0 if the recovered address matches the owner.
+            // Else returns 1, which is equivalent to:
+            // `(sigFailed ? 1 : 0) | (uint256(validUntil) << 160) | (uint256(validAfter) << (160 + 48))`.
+            // where `validUntil` and `validAfter` are 0.
+            validationData := iszero(iszero(sigFailed))
```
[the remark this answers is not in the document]
Vectorized: Yes. I added the `iszero(iszero(x))` for extra-futureproof-safety. 

The compiler optimizes the double-iszero away with and without ir.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,295 @@
...
+        payable
+        virtual
+        onlyEntryPointOrOwner
+    {
+        /// @solidity memory-safe-assembly
+        assembly {
+            let m := mload(0x40)
+            calldatacopy(m, data.offset, data.length)
+            if iszero(call(gas(), target, value, m, data.length, codesize(), 0x00)) {
+                // Bubble up the revert if the delegatecall reverts.
```
kopy-kat, line 157: I think this should be call not delegatecall (same below)

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,295 @@
...
+    /// it will not be required to send again)
+    ///
+    /// `missingAccountFunds` is the minimum value this method should send the EntryPoint,
+    /// which MAY be zero, in case there is enough deposit, or the userOp has a paymaster.
+    function _payPrefund(uint256 missingAccountFunds) internal virtual {
+        /// @solidity memory-safe-assembly
+        assembly {
+            if missingAccountFunds {
+                // Ignore failure (its EntryPoint's job to verify, not the account's).
+                pop(call(gas(), caller(), missingAccountFunds, codesize(), 0x00, codesize(), 0x00))
```
kopy-kat, line 112: Why is `codesize()` used here - wouldnt it be cheaper to just use `0x00` since no calldata is passed and returndata is ignored?

test/ERC4337.t.sol
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    function withdrawTo(address to, uint256 amount) public payable {
+        balanceOf[msg.sender] -= amount;
+        (bool success,) = payable(to).call{value: amount}("");
+        require(success);
+    }
+}
+
+contract ERC4337Test is SoladyTest {
```
kopy-kat, line 38: There's no test coverage for `validateUserOp` and it could also make sense to add an e2e test using the EntryPoint but these would be more for completeness sake

src/accounts/ERC4337.sol
Vectorized: `codesize` costs 2 gas. 0x00 sometimes costs 3 gas. This is the best we can do without resorting to PUSH0, which isn’t widely available.

test/ERC4337.t.sol
Vectorized: Can add it if once recharged energy.

src/accounts/ERC4337.sol
Vectorized: Good find.

test/ERC4337.t.sol
atarapra: @kopy-kat I will working on it and will push more tests.

src/accounts/ERC4337.sol
kopy-kat: ah yes - I misread and thought it was used for `argSize` not offset

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,297 @@
...
+        payable
+        virtual
+        onlyEntryPointOrOwner
+    {
+        /// @solidity memory-safe-assembly
+        assembly {
+            let m := mload(0x40)
+            calldatacopy(m, data.offset, data.length)
+            if iszero(call(gas(), target, value, m, data.length, codesize(), 0x00)) {
+                // Bubble up the revert if the call reverts.
```
z0r0z, line 159: Why not include return data from successful calls? I would argue that revert data is not that useful and could even just make the `if (call(.....` as condition to return success data and not bother with bubbling the revert message.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,297 @@
...
+                returndatacopy(m, 0x00, returndatasize())
+                revert(m, returndatasize())
+            }
+        }
+    }
+
+    /// @dev Execute a sequence of call operations from this account.
+    /// For efficiency, the `values` can be an empty array,
+    /// if not the same length as `targets` and `data`.
+    function executeBatch(
```
z0r0z, line 169: Push batch ops to Multicaller.sol? https://github.com/Vectorized/multicaller/blob/main/src/Multicaller.sol

I expect this would encourage adding a `delegatecall` function to Account.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,297 @@
...
+        address[] calldata targets,
+        uint256[] calldata values,
+        bytes[] calldata data
+    ) public payable virtual onlyEntryPointOrOwner {
+        /// @solidity memory-safe-assembly
+        assembly {
```
z0r0z, line 175: Would a struct for excutable ops be more efficient than length-matching checks?

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,297 @@
...
+        }
+    }
+
+    /// @dev Requires that the caller is the EntryPoint, or the owner, or the contract itself.
+    modifier onlyEntryPointOrOwner() virtual {
+        if (msg.sender != entryPoint()) _checkOwner();
+        _;
+    }
+
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
```
z0r0z, line 220: Since the `balance` and `deposit` are open functions on entryPoint and withdrawals can be called through execute by account, it would be simpler and adequate to ERC4337 to exclude this section izo (inz0r0zopinion).

src/accounts/ERC4337.sol
Vectorized: I think most people don't even know the existence of Multicaller.sol. 

Would like to keep both codebases independent.

src/accounts/ERC4337.sol
Vectorized: I initially wanted to exclude them, but I think that being able to have those stuff show up on Etherscan might be worth the convenience. The function dispatch will be slightly more costly, but it's ok.

src/accounts/ERC4337.sol
Vectorized: I guess we may be able to add a `delegateExecute(address implementation, bytes calldata data)` function. 

Can be quite the footgun, but the only storage slot in the Account right now is the owner slot...

src/accounts/ERC4337.sol
z0r0z: would it make sense to only copy data into a return and ignore or 0x00 out revert case, or, could I nudge just `pop` the call and not revert or return 😆

src/accounts/ERC4337.sol
z0r0z: > I think most people don't even know the existence of Multicaller.sol.
> 
> Would like to keep both codebases independent.

Understand independence code-wise, but could just be in this case, acknowledging or commenting on external batching as option for `execute` route. E.g., `Multicaller` is in `Multisend` lane, and among account users, external batching is fairly common: https://etherscan.io/address/0x40a2accbd92bca938b02010e17a5b8929b49130d

I do see the comparison to SimpleAccount, but this is worth flagging izo.

z0r0z: > I guess we may be able to add a `delegateExecute(address implementation, bytes calldata data)` function.
> 
> Can be quite the footgun, but the only storage slot in the Account right now is the owner slot...

I like including a delegatecall method as light path to upgradeability and allowing for external batching. Footgun as it is, I tend to just see the situation as ... if owner already messed up target `to`, it doesn't matter if it's a call or delegatecall, and same goes to other advanced or implicit risks, such as accidentally sending all `value` out.

src/accounts/ERC4337.sol
z0r0z: ooo nice ok 🍰

src/accounts/ERC4337.sol
Vectorized: Bubbling up reverts is more useful here since it is ultra-cheap and can provide helpful traces in stuff like Tenderly if the tx actually fails.

Collecting returndata isn't that useful, since these execute functions are meant to be mostly called by the EntryPoint, which discards the returndata.

If we're gonna collect and return returndata, we'll need to do the the direct-return style like in https://github.com/Vectorized/solady/blob/main/src/utils/Multicallable.sol or else, it will be super bloat. Need to add those pesky comments for 3 functions.

src/accounts/ERC4337.sol
z0r0z: Design-wise, I get it, and as well the fact that this contract takes L2 more into account (zip fallback). My suggestion though on the face of 'easier on etherscan' is that the demographic for account abstraction won't even know about blockchain explorers but will be collecting magic fruits on some metaversal island and so prob notice more if they have less fruit end'o'day to pay network taxman. Or something like that.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,322 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+import {Receiver} from "./Receiver.sol";
+import {Ownable} from "../auth/Ownable.sol";
```
z0r0z, line 5: I like the promise of flexibility here (e.g., an eoa transfers ownership to a multisig for social recovery or joint account). But, worth commenting: Since the account is already upgradeable, wouldn't it be cheaper and safer for delegatecall-purposes as well, to put owner onto stack as immutable, and then ownership changes are managed via upgrades? My sense is that these would be less frequent or that the init owner would be a contract with flexibility paths.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,322 @@
...
+
+    /// @dev The lengths of the input arrays are not the same.
+    error ArrayLengthsMismatch();
+
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                        INITIALIZER                         */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Initializes the account with the owner. Can only be called once.
+    function initialize(address newOwner) public virtual {
```
z0r0z, line 51: this should also be `payable` izo as one-off function that resembles current thinking on using payable optimizations for `onlyOwner` functions, plus *no-harm-no-foul* coda (any eth paid here likely from owner anyways and therefore recoverable, as well, this allows factories to attach value for init).

Vectorized: good catch

z0r0z:
> supportsInterface(bytes4) (do people actually use this on-chain?)

Re: ERC165 introspection support in 4337 account itself, it's not required by safe token transfer receivers (ERC721, ERC1155), ERC4337 or ERC1271, so let's not bother with adding. Similarly, I haven't ever seen a contract check for this anyways.

z0r0z:
Sidenote: Supporting ERC1271 `isValidSignature()` method makes sense here. Accounts should be able to spend the signature of owner. On that front, would it make sense to put this method into own file, like `Receiver.sol`, or is it so opinionated to ownership that it should be framed into account itself. If Solady perspective is that Ownable == Single Owner Auth, then I think separate file would be most useful for library purposes.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,323 @@
...
+        if (msg.sender != entryPoint()) revert Unauthorized();
+        _;
+    }
+
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                    EXECUTION OPERATIONS                    */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Execute a call operation from this account.
+    function execute(address target, uint256 value, bytes calldata data)
```
z0r0z, line 148: consider: use `value` to switch on `delegatecall` if set to uint256 max (never going to be that much ether or alt gas so it's safe). This would help with things like EIP-6900 which specifies `execute(address,uint,bytes)` but has no delegation method: https://eips.ethereum.org/EIPS/eip-6900

Likewise, if including batch, might as well conform to this EIP.

e.g. `if eq(value, not(0) ....`

Vectorized: Oh shyt... i just realized i really don't like 6900's interfaces.

Vectorized: I am maxi of `executeBatch(address[],uint256[],bytes[])`. 

Structs a PITA to access in assembly. 

@adamegyed

Vectorized: Mmm... i see. Very good!

Ok, I think it's worth to use structs then. Getting the `data` is the pain, but we can remove the need to do the array length check. :)

Vectorized: Just to confirm, `Execution` is defined as:

```solidity
struct Execution {
    address target;
    uint256 value;
    bytes data
}
```

z0r0z: @adamegyed while I have you - is there a reason to include batch execute as requirement in 6900? I see reference to this being added because eth infinitism added this helper to their SimpleAccount sample, but wouldn't an authorized multicall function (common module) or multisend (external helper, of which there are a couple canon examples) be sufficient?

Vectorized: `Call` sounds short and sweet imo.

z0r0z: h `Op` gets validated and executed .  ̫.

z0r0z: @adamegyed helpful to know ~~ multicall might be more familiar path for 6900 and would cover batching any other account operation. Another consideration is relaxing batch func req. if `delegatecall` is implemented. Will make notes in magician's thread to add suggestion.

src/accounts/ERC4337.sol
Vectorized: I think making owner immutable will make it very difficult to verify. The factory will also not be able to use a minimal ERC1967 clone.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,345 @@
...
+        virtual
+        returns (bytes4 result)
+    {
+        bool success = SignatureCheckerLib.isValidSignatureNowCalldata(
+            owner(), SignatureCheckerLib.toEthSignedMessageHash(hash), signature
+        );
+        /// @solidity memory-safe-assembly
+        assembly {
+            // `success ? bytes4(keccak256("isValidSignature(bytes32,bytes)")) : 0xffffffff`.
+            result := shl(224, or(0x1626ba7e, sub(0, iszero(success))))
```
z0r0z, line 106: should we bother with the formatted error selector `0xffffffff` and spend gas when not required by ERC1271? Anything that doesn't return magic value will just default to `0x00000000` empty value anyways.

src/accounts/ERC4337.sol
z0r0z: Helpfully more lately CWIA proxies can be accessed on etherscan through the `is this a proxy` tab, and their implementation can be verified. A benefit though of transferrable ownership though can be a token-bound or ERC6551-like layer on top of this that does `_setOwner` if owner token is sent. Closing comment for now but will look at verification options.

src/accounts/ERC4337.sol
Vectorized: You've read my mind.

Vectorized: I think it's ok. I don't want some people to say that a future version of solc may make `success` become not 1 or 0 under-the-hood. This formulation uses an `iszero` to make it safe.

Might use the shorter `shl(224, mul(success, 0x1626ba7e))` if solc actually adds Solady into their CI.

z0r0z: tbh 0xffffffff just seems like something that carried over from the EIP reference implementation, but when we look at 721 and other bytes4 selector-based returns, they either give a magic value, or they don't. Curious on why it should be there more than anything.

shrugs: magic value only required for the success case, so i think you're good here to return whatever otherwise. all EIP1271 impl check magic value, and i don't know/remember why the reference uses 0xff..

src/utils/SignatureCheckerLib.sol
```diff
@@ -439,6 +439,51 @@ library SignatureCheckerLib {
         }
```
z0r0z, line 439: some potential real estate in SignatureCheckerLib ...
* Do we need to 'clean the upper 96 bits of `signer`' since account checks `owner` and this will be trusted

O/wise, is `iszero(returndatasize())` needed? The recovered address either matches owner or not so could limit extent of check (also applies in the 1271 case?), https://github.com/Vectorized/solady/blob/ab100a595fe2beaf831dc7153dd043d744995e3d/src/utils/SignatureCheckerLib.sol#L117C34-L117C58

Vectorized: The library still needs to clean `signer`.

z0r0z: for sure ~~ but more so wondering thoughts on this case. The other imports are just as optimized regardless of 4337 context but in sig check case the assumptions introduce more gas. Readability so far is good but worth flag izo.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,370 @@
...
+            mstore(0x40, add(o, returndatasize())) // Allocate the memory.
+        }
+    }
+
+    /// @dev Guard to ensure that the owner slot's value isn't changed by the function.
+    /// You can override this modifier to ensure the sanctity of other storage slots too.
+    modifier storageGuard() virtual {
+        bytes32 ownerSlotValue = _ownerSlotValue();
+        _;
+        assert(_ownerSlotValue() == ownerSlotValue);
```
Vectorized, line 261: `require` refunds all remaining gas.

`assert` burns all remaining gas. And doesn't abuse the `returndatacopy(returndatasize(), returndatasize(), mustBeFalse)` trick.

Prefer `require`.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,370 @@
...
+            result :=
+                mul(
+                    mload(0x20),
+                    and( // The arguments of `and` are evaluated from right to left.
+                        gt(returndatasize(), 0x1f), // At least 32 bytes returned.
+                        staticcall(gas(), ep, 0x10, 0x24, 0x20, 0x20)
+                    )
+                )
+        }
+    }
```
Vectorized, line 316: If this trick is used, we must add a comment saying that this function will return zero instead of reverting if the `EntryPoint` doesn't exist.

src/accounts/ERC4337.sol
```diff
@@ -0,0 +1,370 @@
...
+    function addDeposit() public payable virtual {
+        address ep = entryPoint();
+        /// @solidity memory-safe-assembly
+        assembly {
+            if iszero(call(gas(), ep, callvalue(), codesize(), 0x00, codesize(), 0x00)) {
+                mstore(0x00, 0xb12d13eb) // `ETHTransferFailed()`.
+                revert(0x1c, 0x04)
+            }
+        }
+    }
```
Vectorized, line 328: Must go through `depositTo(address to)`.

Vectorized: See:
```solidity
    /**
     * add to the deposit of the given account
     */
    function depositTo(address account) public payable {
        _incrementDeposit(account, msg.value);
        DepositInfo storage info = deposits[account];
        emit Deposited(account, info.deposit);
    }
    ```

Comment (z0r0z) [reply src/accounts/ERC4337.sol:328]: ? the entry point treats all pure eth receipts as a `depositTo(msg.sender)` https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/core/StakeManager.sol#L45C5-L47C6

File: src/accounts/ERC4337.sol (hunk 16 above)
Comment (Vectorized) [reply src/accounts/ERC4337.sol:316]: Actually, I prefer reverting if the `entryPoint` doesn't exist. It only eats a few gas more during runtime. If someone truly wants to save the gas, they will call the `entryPoint.depositTo` directly.

Comment (z0r0z) [reply src/accounts/ERC4337.sol:316]: oki, but something I think might just be assumed since the implementation here hardcodes the current canon `EntryPoint`, is that it's a trusted call to deployed code. The prefund modifier function for example does a simple popped ether call, and the EP will have admin priveleges by default. It might therefore be reasonable to not just use this trick, but do something simpler, like this:
```solidity
    function getDeposit() public view virtual returns (uint256) {
        address ep = entryPoint();
        /// @solidity memory-safe-assembly
        assembly {
            mstore(0x14, address()) // Store the `account` argument.
            mstore(0x00, 0x70a08231000000000000000000000000) // `balanceOf(address)`.
            pop(staticcall(gas(), ep, 0x10, 0x24, 0x01, 0x20))
            return(0x01, 0x20)
        }
    }
```

z0r0z: likewise (again, the thought is that the EP is a core assumption to account working at all and just following singletons): 
```solidity
function addDeposit() public payable virtual {
        address ep = entryPoint();
        /// @solidity memory-safe-assembly
        assembly {
            pop(call(gas(), ep, callvalue(), codesize(), 0x00, codesize(), 0x00))
        }
    }
```

PR #644, opened 2023-10-14
✨ ERC4337Factory

Vectorized:
## Description

Help me write tests.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


atarapra:
Update : Removed `deployAndCall` and `deployDeterministicAndCall` function because in `ERC4337` accounts most off function are only callable by `owner` or `entryPoint`. So, factory are not able call this functions.

atarapra:
@z0r0z big thanks. Forgot to remove `_call` function.

z0r0z:
np @atarapra I also agree on your simplification here. The most that a factory call could do is fwd a deposit, but even that case can be covered via init userOps.

src/accounts/ERC4337Factory.sol
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /// @dev Deploys an ERC4337 account with `salt` and returns its deterministic address.
+    function deployDeterministic(address owner, bytes32 salt)
+        public
+        payable
+        virtual
+        returns (address account)
+    {
+        LibClone.checkStartsWithCaller(salt);
```
z0r0z, line 48: I think we can remove this 'safe' create2 check @Vectorized. Since EntryPoint will often be the factory caller, account salts can still be frontrun if observed and similarly fwd'd as userOp with EP in first 20 bytes.

Vectorized: Wdyt about `checkStartsWithCaller(bytes32 salt)` -> `checkStartsWith(bytes32 salt, address by)`.

Then `LibClone.checkStartsWith(salt, owner)`.

z0r0z: That's neat for fix. Yeah I think this prevents anyone from swapping out the owner to steal the create2 account destination's funds while in transit. Also overall cleaner for lib to dissac. relayers from cloning ops. Ps: I think `by` in this new func might just be called `for` in param or like 'grain' or somethin. Behold `salt2` 🧂 🧂

PR #652, opened 2023-10-18
✨ ERC6551 Simple Account

z0r0z:
## Description

Begins a simple ERC6551 account.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
@atarapra 

I think the constructor of ERC6551 should automatically deploy a minimal ERC1967Proxy with the default implementation baked into the bytecode. Similar to the minimal ERC1967 proxy in LibClone, but with the default logic in ERC6551Proxy..

Maybe we can write a function called `LibClone.deployERC1967D` in LibClone.

The minimal proxy's address will then be saved into a public immutable variable in ERC6551.

Then we can delete ERC6551Proxy after that.

Vectorized:
**TIL:**

```
Compiler run successful with warnings:
Warning (3628): This contract has a payable fallback function, but no receive ether function. Consider adding a receive ether function.
 --> src/accounts/ERC6551Proxy.sol:7:1:
  |
7 | contract ERC6551Proxy {
  | ^ (Relevant source part starts here and spans across multiple lines).
Note: The payable fallback function is defined here.
  --> src/accounts/ERC6551Proxy.sol:36:5:
   |
36 |     fallback() external payable {
   |     ^ (Relevant source part starts here and spans across multiple lines).


```

You can shut the compiler up by making `_defaultImplementation` internal.

As long as the contract doesn't have a single non-payable public method, it silences the warning.

**Also:**

```
mstore(0x40, returndatasize()) // This replaces the `PUSH1 0x80` with a `RETURNDATASIZE`.
```

PR #661, opened 2023-10-20
⚡ Optimize ERC6551

z0r0z:
## Description

Pin (feel free to nuke) potential optimizations in LOC and gas consumption.

* 🤏  shorten owner get 
*won't chainid be handled by registry? attack vectors seem remote. If staticcall fails, then result won't be assigned and will be address(0) right so I think this is worth considering.*
* ... // 

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
Will merge this. It's required to bypass a stack too deep in older solc with via-ir.

Vectorized:
I think let's short-circuit based on chain ID. I don't know why, but seems faster. And gut feel says it might be useful for cross-chain stuff as described in the doc.

PR #679, opened 2023-10-25
✨ Add LibPackedArray

haydenshively:
## Description

I've been working on an [implementation of ERC721](https://github.com/haydenshively/solady/pull/1) (enumerable) that utilizes SSTORE2 to save a list of each user's `tokenId`s. To maximize efficiency, the `tokenId`s should be packed if < 32 bytes, and I needed a library for efficient operations on such arrays.

This is my solution. Likely room for further optimization. Tried to stick with JS nomenclature for naming array methods, with the exception of `remove`, for which there was no equivalent.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
I’d probably write a thin wrapper around DynamicBufferLib. Didn’t include this cuz too niche and the name strongly suggests a library that handles packed arrays in storage.

atarapra:
@haydenshively thanks for this PR, and I appreciate your work. I agree with @Vectorized . However, I reviewed your 'ERC721C' contract and don't believe it's very efficient when it comes to 'transfer' and 'mint' operations. I can't identify any major use cases for this library on-chain. Additionally, there seem to be some limitations when selecting the 'chunkSize' for optimal efficiency.

Vectorized:
I think we should just use LibMap if the need for a ERC721Enumerable is really that big. Looking at OZ's implementation, it looks like one can reduce the number of cold SSTOREs by 4? Probably good enough.

My opinionated view is that it is better to just query the `tokenIds` off-chain, and then pass in an array of `tokenIds` owned by the user and verify ownership if needed.

PR #688, opened 2023-10-31
♻️ Use nested EIP-712 approach for accounts

Vectorized:
Many thanks to @junomonster, @frangio, and others in the 1271 group for the ideas and reviews.

Typescript code and demo for the frontend:  
- https://github.com/junomonster/nested-eip-712 
- https://github.com/frangio/eip712-wrapper-for-eip1271 

## Description

The final hash passed into ecrecover will therefore be an EIP-712 hash tied to the contract's address and chain ID via the domain separator `DOMAIN_SEP_A`.

For the nested EIP-712 approach will be, the final hash will be:
```
keccak256(\x19\x01 || DOMAIN_SEP_A ||
    hashStruct(Parent({
        childHash: keccak256(\x19\x01 || DOMAIN_SEP_B || hashStruct(originalStruct)), // Passed in as `hash`.
        child: hashStruct(originalStruct) // Passed in from `signature`.
    }))
)
``` 
where `||` denotes the concatenation operator for bytes.  
The signature will be `r || s || v || PARENT_TYPEHASH || DOMAIN_SEP_B || child`.

The implementation will use `child` and `DOMAIN_SEP_B` from the signature to check if `childHash` is correct.

For the `personal_sign` workflow, the final hash will be:
```
keccak256(\x19\x01 || DOMAIN_SEP_A ||
    hashStruct(Parent({
        childHash: personalSign(someBytes) // Passed in as `hash`.
    }))
)
``` 
where `||` denotes the concatenation operator for bytes.  
The signature will be `r || s || v || PARENT_TYPEHASH`.

## Why?

Ain't nobody got time to wait for Metamask and EIPs.

Anyway, nobody uses Solady accounts yet (according to codeslaw), so it's ok if everyone else prefers doing it their own way.

I'm simply suggesting the most tasteful way in my opinion. ;) 

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/accounts/ERC6551.sol
```diff
@@ -131,30 +122,59 @@ abstract contract ERC6551 is UUPSUpgradeable, Receiver, EIP712 {
         virtual
         returns (bytes4 result)
     {
-        bool success = SignatureCheckerLib.isValidSignatureNowCalldata(
-            owner(), _computeIsValidSignatureDigest(hash), signature
-        );
+        bool success = _isValidSignatureWithNestedEIP712(hash, signature);
```
junomonster, line 125: quick question:
The `innerHash == hash(childData)` check would be in the caller contract, right?

Vectorized: Just added a check for `innerHash == hash(childData)`.

Vectorized: Of course, the perfect solution is that every wallet UI maker listen to my beck and call. But we live in a society.

So let's aim for the solution closest to perfection with only smart account implementation and frontend code.

frangio:
It's better to use compact `r, vs` representation to keep the signature a multiple of 32 and avoid 0 when ABI encoded.

Vectorized:
Will probably let it marinate for about 3 days before merge.

PR #689, opened 2023-11-01
✨ Add EIP-2098 fallbacks

Vectorized:
## Description

Everyone (who uses Solady) uses nonces. I think.

With the spread of smart wallets on L2s, which use `bytes` for signature due to 4337 and 1271, this makes sense.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
Thanks @Philogy for the re-review.

src/utils/ECDSA.sol
```diff
@@ -36,17 +36,30 @@ library ECDSA {
     function recover(bytes32 hash, bytes memory signature) internal view returns (address result) {
```
Philogy, line 36: You need to update the comment.

Philogy:
Looks good to me, just some small notes:
- you need to update the comments on the 4 updated methods that EIP-2098 is now accepted
- I'd switch the normal/EIP-2098 check blocks to optimize for EIP-2098, in practice I think this is likely to be the more commonly used one

Generally I'm not sure if it's a good idea to add both EIP-2098 and normal signature checking code for the same methods, I'd create separate variants that allows devs to choose, allowing each to be more compact and gas efficient.

PR #704, opened 2023-11-04
📝 Intro to WETH docs

tolu0x:
I wrote a short intro to WETH's doc. I'm curious as to what the general style guide for solady's documentation is. Lmk if i can continue like this or there are things I should change. @Vectorized


Vectorized:
Solady's WETH intended purpose is to provide an example for testing and educational purposes. 

For serious business, users should consider [METH](https://github.com/Philogy/meth-weth) which is more complete and efficient.

PR #734, opened 2023-11-25
✨ Add LEB128Lib

Shungy:
## Description

Add  library to encode and decode values based on LEB128 compression scheme. Useful for non-standard way of encoding calldata to decrease calldata size. See the [usage example](https://github.com/Shungy/leb128-nooffset/blob/14f6ae06d39a95733a60ebcc2a62c3cd2e8ac80b/src/examples/EncodedCalldataERC20.sol).

## Checklist

- [✔ ] Ran `forge fmt`?
- [✔] Ran `forge snapshot`?
- [✔] Ran `forge test`?


atarapra:
@Shungy how much this method is efficient in compare to `LibZip.cdCompress()` and `LibZip.flzCompress()` methods.

Vectorized:
Might not accept.

Tried it on this sample calldata:
```
0x4a04a1c90000000000000000000000000000000000000000000000000000000000000020000000000000000000000000ca1694433e499862ee242f2f403cb1e73ae91cfb000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000001f5d295778796a8b9f29600a585ab73d452acb1c0000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000ffffffff0000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000026000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

which, broken up into uint256s, will give:
```
function testBenchBE128Compress() public {
        uint256[] memory a = new uint256[](21);
        a[0] = 0x0000000000000000000000000000000000000000000000000000000000000020;
        a[1] = 0x000000000000000000000000ca1694433e499862ee242f2f403cb1e73ae91cfb;
        a[2] = 0x0000000000000000000000000000000000000000000000000000000000000001;
        a[3] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[4] = 0x0000000000000000000000001f5d295778796a8b9f29600a585ab73d452acb1c;
        a[5] = 0x0000000000000000000000000000000000000000000000000000000000000001;
        a[6] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[7] = 0x00000000000000000000000000000000000000000000000000000000ffffffff;
        a[8] = 0x0000000000000000000000000000000000000000000000000000000000000200;
        a[9] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[10] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[11] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[12] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[13] = 0x0000000000000000000000000000000000000000000000000000000000000220;
        a[14] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[15] = 0x0000000000000000000000000000000000000000000000000000000000000260;
        a[16] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[17] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[18] = 0x0000000000000000000000000000000000000000000000000000000000000014;
        a[19] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        a[20] = 0x0000000000000000000000000000000000000000000000000000000000000000;
        uint256 total = 0;
        for (uint256 i; i < a.length; ++i) {
            total += LEB128Lib.encode(a[i]).length; 
            total += 1; // We need at least 1 byte denote how long the bytes is.
        }
        console.log(total);
    }
```

This gives a total of `93 bytes + 4 function signature bytes = 97 bytes`. 

`cdCompress` gives a total of `79 bytes`.

Shungy:
> `             total += 1; // We need at least 1 byte denote how long the bytes is.`

In LEB128 the length is already encoded in the value. So if you are decoding all your calldata at once then there is absolutely no need to define lenghts or offsets for values. So `97` becomes `76`, beating `cdCompress`.

Added benchmark to the tests. Basically this is possible:

```
        bytes memory encodedData;
        for (uint256 i; i < a.length; ++i) {
            encodedData = abi.encodePacked(encodedData, LEB128Lib.encode(a[i]));
        }
        assertEq(encodedData.length, 72);

        // Test that we can decode these values.
        uint256 ptr;
        assembly {
            ptr := add(encodedData, 0x20)
        }
        for (uint256 i; i < a.length; ++i) {
            uint256 result;
            (result, ptr) = LEB128Lib.rawMemDecodeUint(ptr);
            assertEq(result, a[i]);
        }
```

Shungy:
> @Shungy how much this method is efficient in compare to `LibZip.cdCompress()` and `LibZip.flzCompress()` methods.

I haven't tested. I guess it depends on the data as this is a different method of compression (?). But LEB128 length can be easily determined based on the most significant bit position of the value: https://github.com/Shungy/solady/blob/d75459dab1ae8c31664070ec92f6c558a46bd9d0/test/LEB128Lib.t.sol#L68-L75

Vectorized:
@Shungy Can you write a PoC to demonstrate using BE128 on any calldata?

Shungy:
Very simple example: https://github.com/Shungy/leb128-nooffset/blob/14f6ae06d39a95733a60ebcc2a62c3cd2e8ac80b/src/examples/EncodedCalldataERC20.sol#L31-L36

It's not plug and play.

I can write a PoC for complex calldata encoding as well but that's beyond the scope of this library, as there can be different ways to standardize that based on how you want to decode  non integer values.

Vectorized:
Lemme think about it. For specific cases, LEB might offer slightly better compression. 

I have another example where the inputs don't lend well to simply LEB every 32 byte chunk.

Shungy:
> I have another example where the inputs don't lend well to simply LEB every 32 byte chunk.

Yea that wasn't my intended way of using. If you're encoding right padded 32 bytes then the encoded value will be 36-37 bytes.

Vectorized:
My OCD is slightly itching over having two separate compression sol files. 

LEB definitely need more SLOC, so copypastaing it into LibZip will bulk up the file.

Vectorized:
After much thought, I think it’s best to just stick with LibZip for Solady. The best case of LEB128 can save a few bytes, but it requires more work (on both solidity side and client side) and is not general.

JFC, it took quite a while for sound to integrate LibZip cdCompress, even tho there’s already a JS plug and play solution.

On average, LEB128 use 22.86 bytes for an address. This is similar to cdCompress’ usage of 22 bytes per address. The case where LEB shines is where there are many small numbers in the calldata. 

I greatly appreciate the work and thought put into this. I’m very sorry. 🙇‍♂️

Shungy:
Thanks for looking into it :pray: :pray:

PR #751, opened 2023-12-14
🥢 Fix lnWad comments

xenoliss:
## Description

Was working on porting the `FixedPointMathLib` for int128 and spent quite some time figuring out how some magic numbers were computed. Ended out that the comments were slightly wrong, so fixing this might help others as well.

**PS**: I'm facing a challenge with the implementations of `lnWad` and `expWad`. The documentation states that *'Care is taken to make sure ln(1) is exactly 0'* and *'Care is taken to make sure that exp(0) is exactly 1'*. However, I am struggling to achieve these exact values in int128. I would appreciate any guidance or insights you might have on this issue.

Thank you for developing this awesome lib!

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?

### DIFF
--- a/src/utils/FixedPointMathLib.sol
+++ b/src/utils/FixedPointMathLib.sol
@@ -275,12 +275,12 @@ library FixedPointMathLib {
             // The q polynomial is known not to have zeros in the domain.
             // No scaling required because p is already `2**96` too large.
             r := sdiv(p, q)
-            // Multiply by the scaling factor: `s * 5e18 * 2**96`, base is now `5**18 * 2**192`.
+            // Multiply by the scaling factor: `s * 5**18 * 2**96`, base is now `5**18 * 2**192`.
             r := mul(1677202110996718588342820967067443963516166, r)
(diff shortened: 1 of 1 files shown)


Vectorized:
The magic numbers are created by @remco.

https://2π.com/22/exp-ln/ <- Theory, but note that the python notebook linked in this article can't produce the exact values.

Vectorized:
Lemme merge this first.

xenoliss:
> If you are not too concerned with gas, just upcast to int256, call lnWad or expWad, and then downcast it back to int128.

Oh yes I would definitely do that but I'm porting this to non EVM languages so basically need to reimplement the algorithm from scratch.

> [https://2π.com/22/exp-ln/](https://xn--2-umb.com/22/exp-ln/) <- Theory, but note that the python notebook linked in this article can't produce the exact values.

I've seen this article, but I didn't know about the playbook. I'll check it out, thanks!

PR #764, opened 2023-12-18
🐞 Fix Javascript LibZip.flzCompress

mdehoog:
## Description

I noticed the Javascript implementation of `LibZip.flzCompress` doesn't quite match the solidity version, which results in some different output on certain inputs.

Ths PR fixes the compression method to exactly match the output of the Solidity version (at least, in all of the inputs I've tested).

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?

### DIFF
--- a/js/solady.js
+++ b/js/solady.js
@@ -95,10 +95,10 @@
      */
     LibZip.flzCompress = function(data) {
         var ib = hexToBytes(hexString(data)), b = ib.length - 4;
-        var ht = [], ob = [], a = 0, i = 2, o = 0, j, s, h, d, c, l, r, p, q, e, m = 0xffffff;
+        var ht = [], ob = [], a = 0, i = 2, o = 0, j, s, h, d, c, l, r, p, q, e;
 
-        function u32(i) {
-            return ib[i] | (ib[++i] << 8) | (ib[++i] << 16) | (ib[++i] << 24);
+        function u24(i) {
+            return ib[i] | (ib[++i] << 8) | (ib[++i] << 16);
         }
 
         function hash(x) {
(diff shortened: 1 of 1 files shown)


js/solady.js
```diff
@@ -112,17 +112,17 @@
 
         while (i < b - 9) {
             do {
-                r = ht[h = hash(s = u32(i) & m)];
-                c = (d = (ht[h] = i) - r) < 8192 ? u32(r) & m : m + 1;
+                r = ht[h = hash(s = u24(i))] || 0;
```
mdehoog, line 115: this is one of the bugs: reading `ht` at an index past it's length would return `undefined`, which broke the next line

js/solady.js
```diff
@@ -112,17 +112,17 @@
...
             if (i >= b - 9) break;
             if (--i > a) literals(i - a, a);
             for (l = 0, p = r + 3, q = i + 3, e = b - q; l < e; l++) e *= ib[p + l] === ib[q + l];
-            s = u32(i += l);
+            i += l;
             for (--d; l > 262; l -= 262) ob[o++] = 224 + (d >> 8), ob[o++] = 253, ob[o++] = d & 255;
             if (l < 7) ob[o++] = (l << 5) + (d >> 8), ob[o++] = d & 255;
             else ob[o++] = 224 + (d >> 8), ob[o++] = l - 7, ob[o++] = d & 255;
-            ht[hash(s & m)] = i++, ht[hash(s >> 8)] = i++, a = i;
+            ht[hash(u24(i))] = i++, ht[hash(u24(i))] = i++, a = i;
```
mdehoog, line 125: second bug is here... in the solidity version, `i` is used as the input to the hash function, rather than `s`. `s` === `i` the first time, but not the second time.

js/solady.js
```diff
@@ -95,10 +95,10 @@
      */
     LibZip.flzCompress = function(data) {
         var ib = hexToBytes(hexString(data)), b = ib.length - 4;
-        var ht = [], ob = [], a = 0, i = 2, o = 0, j, s, h, d, c, l, r, p, q, e, m = 0xffffff;
+        var ht = [], ob = [], a = 0, i = 2, o = 0, j, s, h, d, c, l, r, p, q, e;
 
-        function u32(i) {
-            return ib[i] | (ib[++i] << 8) | (ib[++i] << 16) | (ib[++i] << 24);
+        function u24(i) {
```
mdehoog, line 100: switching this to `u24` allowed us to remove the `0xffffff` mask on all of the `u32` calls

Vectorized:
@mdehoog Thank you so much!

Vectorized:
Ok, i differential fuzzed with hardhat.

Vectorized:
Impact:

Previously, the JS compressed data will still decompress correctly into the original data via JS or Solidity. 

But the compression isn't as optimized as it should be.

PR #772, opened 2023-12-29
♻️ Chore: Update ERC6909 Metadata Ext.

z0r0z:
## Description

Update `ERC6909` implementation and test files to reflect finalized [`IERC6909Metadata`](https://github.com/jtriley-eth/ERC-6909/blob/main/src/interfaces/IERC6909Metadata.sol).

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


z0r0z:
Also, in other nit: might as well add `contractURI()` getter from [`IERC6909ContentURI`](https://github.com/jtriley-eth/ERC-6909/blob/main/src/interfaces/IERC6909ContentURI.sol#L11) since the current Solady implementation includes the `tokenURI(uint256. id)` method. So the implementation could fully support both metadata and content extensions out-of-box with just that one func addition.

I am catching up on the deliberations around ERC6909 (I like the slimmed down transfers 📏 ). But my ideal would actually be that ERC6909 meta/content extensions would defer things like name/symbol/decimals/metadata to the `tokenURI` pointer and endorse a json format or something. That would save some gas for logic that needs to run onchain. Curious thoughts @jtriley-eth 🖖

Vectorized:
LGTM. Merging in a while.

PR #785, opened 2024-01-09
♻️ ReentrancyGuard: avoid using magic numbers and initialize slot inside of constructor

zerosnacks:
## Description

Initializes `ReentrancyGuard` to non-zero value in constructor to move cost to deployer, not first caller.

Introduces `_NOT_ENTERED` and `_ENTERED` flags to avoid magic numbers - this should come at no additional gas cost.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
Hmm... 2 and 1 are very obvious tho.

Vectorized:
Hmm... this saves gas for the first minter, but the total amount of gas used by the deployer and first minter will be higher. Over the super long term, users of Solady (deployers + users) will be burning more gas overall. Not sure if the extra bytecode is worth it if people are trying to sneak under the spurious dragon limit.

Vectorized:
May revert.

zerosnacks:
> May revert.

No worries! Your reasoning makes sense to me. Libraries ultimately make different trade-offs than implementations.

PR #795, opened 2024-01-12
✨ Allow initializable slot to be overridden

Vectorized:
## Description

@howydev 

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/utils/Initializable.sol
```diff
@@ -53,30 +58,29 @@ abstract contract Initializable {
...
+            // Set `initializing` to 1, `initializedVersion` to 1.
+            sstore(s, 3)
             // If `!(initializing == 0 && initializedVersion == 0)`.
             if i {
                 // If `!(address(this).code.length == 0 && initializedVersion == 1)`.
                 if iszero(lt(extcodesize(address()), eq(shr(1, i), 1))) {
                     mstore(0x00, 0xf92ee8a9) // `InvalidInitialization()`.
                     revert(0x1c, 0x04)
                 }
+                s := shl(shl(255, i), s) // Skip initializing if `initializing == 1`.
```
howydev, line 75: could move this into the if block of L81

src/utils/Initializable.sol
```diff
@@ -108,7 +112,7 @@ abstract contract Initializable {
         /// @solidity memory-safe-assembly
         assembly {
             // Set `initializing` to 0, `initializedVersion` to `version`.
-            sstore(_INITIALIZABLE_SLOT, shl(1, version))
+            sstore(s, shl(1, version))
```
howydev, line 115: (curious) does caching `shl(1, version)` do anything here for gas, or is the compiler smart enough to automatically do a dup under the hood?

howydev:
Thanks for this change senpai! 1 observation - for OZ's initializable, a dev can use the other 23 bytes (bytes32 - uint64 - bool = bytes23) rather safely, but the storage writes here would wipe the data. 99.9% of use cases would never touch those bytes but still suggest to document this somewhere

atarpara:
> Thanks for this change senpai! 1 observation - for OZ's initializable, a dev can use the other 23 bytes (bytes32 - uint64 - bool = bytes23) rather safely, but the storage writes here would wipe the data. 99.9% of use cases would never touch those bytes but still suggest to document this somewhere

Openzeppelin is using struct for these variables so for using left byte23 dev need to add variables into [InitializableStorage](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/utils/Initializable.sol#L65C12-L65C32) struct unless complier will not able to packed up variables.

atarpara:
@Vectorized LGTM!

src/utils/Initializable.sol
Vectorized: I think just recalculate. I am trying to reduce the number of shared variables between the before and after blocks around the `_`, which helps to reduce stack-too-deep.

PR #800, opened 2024-01-22
✨ Add _mintAndSetExtraDataUnchecked in ERC721

### LINKED ISSUE #796: ⚡️ERC721: Combine _mint() and set _setExtraData() into single SSTORE
Currently, the ownership slot can be extended with `extraData` to make use of the extra bits. I've used this feature extensively and it leads to gas savings due to re-using the same slot, which is great.

However, the functions are different and and SSTORE operation happens separately, causing more gas the be used than necessary.

Everytime I've used _setExtraData, it's been right after a _mint, and this seems like a very common pattern from what I've seen others do.

It would be great if there was a single function, `_mintAndSetExtraData`, that does a single SSTORE instead.

atarpara:
## Description

close #796.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized (changes requested):
Let's change it to `_mintAndSetExtraDataUnchecked`.

- No zero address check. 
- No already exists check.

Vectorized (approved):

PR #827, opened 2024-02-22
⚡Optimize MinHeapMemLib

atarpara:
## Description

Optimized  #826
 
1. Removed `capacity` variable 
2. Decrease Stack variable
3. Cache Memory Calculations
4. Reduced code size

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
We have to store the capacity. Otherwise, pushing and popping 1 element will trigger reallocation.

atarpara:
> We have to store the capacity. Otherwise, pushing and popping 1 element will trigger reallocation.

It will happens only when heap size is `1`. I think it is quite rare.

atarpara:
> We have to store the capacity. Otherwise, pushing and popping 1 element will trigger reallocation.

Oops. It will triggered when heap size equals capacity. Thanks a lot for address this issues.

src/utils/MinHeapLib.sol
```diff
@@ -494,7 +493,7 @@ library MinHeapLib {
                 pos := rightPos
                 childPos := add(shl(1, pos), 1)
             }
-            // Siftdown.
+            // Siftup.
```
Vectorized, line 496: Let’s keep the names the same. According to Python’s implementation.

Vectorized (changes requested):

Vectorized (approved):

PR #862, opened 2024-03-20
✨ Coinbase smart wallet ERC1271

Vectorized:
## Description

I'm letting this marinate for a while first. May or may not merge.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/accounts/ERC1271.sol
```diff
@@ -6,7 +6,16 @@ import {SignatureCheckerLib} from "../utils/SignatureCheckerLib.sol";
 
 /// @notice ERC1271 mixin with nested EIP-712 approach.
 /// @author Solady (https://github.com/vectorized/solady/blob/main/src/accounts/ERC1271.sol)
+/// @author Coinbase (https://github.com/coinbase/smart-wallet)
 abstract contract ERC1271 is EIP712 {
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                         CONSTANTS                          */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev `keccak256("CoinbaseSmartWalletMessage(bytes32 hash)")`.
```
wminshew, line 15: is there a composability advantage to using this over something else like `SoladySmartWalletMessage` ?

Vectorized: I used it cuz I saw this in https://github.com/wilsoncusack/scw-tx/blob/ad2323ea56cc3561eb4fed3e6c93fa8354f5a8f8/ERC1271.test.ts#L54. 

The goal is to allow wallets using Solady's 1271 to be compatible with any frontends that is compatible with Coinbase's smart wallets.

I think we can swap out for some more generic sounding name. But would like to keep in consistency with Coinbase's implementation for now.

@wilsoncusack

Vectorized: Some additional background. 

Alchemy light account is using the Solady's default workflow.

https://github.com/alchemyplatform/light-account/blob/develop/src/common/ERC1271.sol

The issue with supporting sign in for smart accounts is that 1271 alone does not standardize replay protection with EIP-712 visibility, so frontends have to potentially support several major smart wallets.

I envision that frontends will eventually congregate to support a few major smart wallets (likely via their SDKs). Solady's 1271 aims to be a drop-in solution to work with these SDKs.

src/accounts/ERC1271.sol
```diff
@@ -81,7 +127,7 @@ abstract contract ERC1271 is EIP712 {
                     hash := keccak256(0x00, 0x60) // Compute the parent's structHash.
                     break
                 }
-                // Else, use the `personal_sign` workflow.
+                // Else, use the `personalSign` workflow.
                 // Truncate the `signature.length` by 1 word (32 bytes), until zero.
```
vigneshka, line 131:
```suggestion
                // Reduce the signature length by 1 word (32 bytes) if it exceeds 1 word, otherwise set it to zero.
```
original comment makes it seem like it repeats some op until it's zero

src/accounts/ERC1271.sol
wilsoncusack: Hey just catching up here. I'm open to making the name more generic or maybe some ERC here. Could just be SmartWalletMessage?

Vectorized:
Closing this for now.

src/accounts/ERC1271.sol
wminshew: @wilsoncusack would be great to have an ERC here I think imo

PR #871, opened 2024-03-27
✨ Lazy storage based Fisher-Yates shuffle

aschlosberg:
## Description

Fisher–Yates shuffle of a lazy-initialised storage array, resulting in a permutation of `[0,n)`. An example use case is shuffling NFTs, amortising gas costs across all mints.

Implementation TL;DR: use a random storage slot as the first word in the array backing and store a value `x` at index `i` as `x==i ? 0 : x+1`, which is trivially invertible such that the default (zero) value at an index is the index itself. This initialises the array to `[0,n)` (in order) by only storing the array slot. The rest is a standard F–Y shuffle.

I chose `uint32` values to strike a balance between (a) something able to fit any feasible array size, and (b) maximising packing to avoid SSTOREs on current-value-zero slots. However I'm leaning towards reducing this to `uint16` as 65k values feels adequate and it halves the number of zero-value overwrites—I'd appreciate other perspectives, please.

## Checklist

- [x] Ran `forge test`?
- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?


Vectorized:
Not sure of the real world use case tho. 

In memory shuffle can handle tokenURI for 99.99% of NFT collections today.

aschlosberg:
> Not sure of the real world use case tho.
> 
> In memory shuffle can handle tokenURI for 99.99% of NFT collections today.

I originally wrote this for a project that required random distribution that could later be used on-chain for token actions dependent on the permutation. Running the first `tokenId` steps of a shuffle each time wouldn't be feasible nor fair to owners with high token IDs.

I won't delete the branch, in case this thread turns up in a Google search and someone wants to use it.

Vectorized:
@aschlosberg 

I see, this is a legit use case.

Lemme reconsider if there is a better algo.

Vectorized:
I'm thinking of moving this into `LibPRNG`.

```solidity
struct LazyShuffler {
    uint192 _slot;
    uint32 length;
    uint32 numShuffled;
}

/// @dev Stores the `slot`, `length`, `numShuffled` into a single slot.
/// The next random number is simply keccak256(state), as `numShuffled` is incremented.
function init(LazyShuffler storage $, uint32 n, uint256 seed) internal;

/// @dev Throws if `i` >= `numShuffled`, or if `$` has not been initialized.
function get(LazyShuffler storage $, uint32 i) internal returns (uint32 shuffledIndex);

/// @dev Throws if the current `numShuffled` >= `length`.
function lazyShuffleNext(LazyShuffleState storage $) internal returns (uint32 result)
```

Each `uint32` chunk will be lazily initialized to `1 + index` as we lazily-shuffle the array.

aschlosberg:
Some comments and questions in-line, below. Let me know what you think and I'll move it into `LibPRNG`. I also thought of a way to remove `set(i, val)` but still be able to test the `set()`–`get()` loop (easier to just implement than to explain, but telling you since I see you removed it in your interface).

> I'm thinking of moving this into `LibPRNG`.
> 
> Having shuffles splitted across two files just feel weird to me.

Makes sense.

> Will link your repo in one of the `@author`s fields.

Thanks :)

> 
> ```solidity
> /// @dev All fields are dangerous to be modified directly.
> /// Maybe we should just combine all into a single `uint256 _state` and make getter methods.
In this case I don't feel strongly either way re getter vs struct fields. I typically lean towards readability (here struct fields) but understand the desire to make unintentional setting more difficult.

> ```solidity
> struct LazyShuffler {
>     uint192 _arraySlot;
>     uint32 length; 
>     uint32 numShuffled;
> }
> 
> /// @dev Stores the `_arraySlot`, `length`, `numShuffled` into a single storage slot.
> /// `_arraySlot` is initialized to `uint192(uint256(keccak256(abi.encode($._arraySlot, $.slot)))`.
What's the rationale behind `($._arraySlot, $.slot)`? My original API was based purely on a user-supplied key for simplicity and clarity (same key => same storage).

> ```solidity
> /// The next random index is simply `keccak256(state) % (length - numShuffled)`,
> /// as `numShuffled` is incremented.

Without the ability to provide an entropy source, the permutation is known in advance, which may not be desirable for some use cases. This is why my code allows for a user-provided random number / seeded PRNG.

> ```
> function init(LazyShuffler storage $, uint32 length, uint256 seed) internal;

Where is `seed` used?

> ```solidity
> /// @dev Returns the `i`th element of the shuffled-portion of array.
> /// Throws if `i` >= `numShuffled`, or if `$` has not been initialized.
> function get(LazyShuffler storage $, uint32 i) internal returns (uint32 shuffledIndex);
> 
> /// @dev Increments `numShuffled`, and returns the next element from `[0..length)`,
> /// as if it has been fully shuffled.
> /// Throws if the current `numShuffled` >= `length` (i.e. no more elements to shuffle).
> function next(LazyShuffleState storage $) internal returns (uint32 result)
> ```
> 
> Each `uint32` chunk will be lazily initialized to `1 + index` as we lazily-shuffle the array.

To `index`, not `1 + index`. Otherwise the permutation is of `1 … n` not `0 … (n-1)`.

> 
> I feel we can conditionally use a `uint16` workflow if `length` is small. Since bulk of gas will be on storage, it's ok to add in a few extra branches.

SGTM

PR #874, opened 2024-04-03
✨ Lazy storage based Fisher-Yates shuffle

Vectorized:
## Description

Continued from #871.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?

### DIFF
--- a/src/utils/SSHUFFLE.sol
+++ b/src/utils/SSHUFFLE.sol
@@ -1,0 +1,143 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {LibPRNG} from "./LibPRNG.sol";
+
+/// @notice Fisher–Yates shuffling of a storage-based, lazy-initialised array.
+/// @author @divergencearran (github.com/aschlosberg)
+/// @custom:source Modified from NextShuffler of github.com/divergencetech/ethier
+library SSHUFFLE {
+    using LibPRNG for LibPRNG.PRNG;
+
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                         CONSTANTS                          */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Bit-width of values; this MUST be the same as the size used as values / size.
+    uint256 internal constant BITS = 32;
+    uint256 internal constant MAX = 0xffffffff;
+
+    /// @dev Number of bits required to shift a value from one end of a word to the other.
+    uint256 internal constant E2E_SHIFT = 224; // 256 - BITS
+
+    /// @dev As it says on the tin.
+    uint256 internal constant LEFT_MASK =
+        0xffffffff00000000000000000000000000000000000000000000000000000000;
+
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                          STRUCTS                           */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev State of a storage-based shuffle. Initialise with `init()`, not directly.
+    /// @dev Instead of using a uint32[] storage, we choose a random slot (see `init()`) and use the
+    /// next `size/8` words to encode a permutation of [0,size). A value of 0 (default) indicates
+    /// an unshuffled index whereas a value of `x+1` indicates that the value `x` has been shuffled
+    /// to the index. As a result, `type(uint32).max` can't be supported; see `set()` and `init()`
+    /// comments.
+    struct State {
+        // Order MUST NOT change.
+        uint192 array; // Slot of first word.
+        uint32 size;
+        uint32 shuffled;
+    }
+
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                         OPERATIONS                         */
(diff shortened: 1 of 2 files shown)


Vectorized:
Letting this marinate a while.

Hiding `get` feels more aesthetically pleasing. But someone might need to read the shuffled array afterwards.

aschlosberg:
> Letting this marinate a while.
> 
> * Hiding `get` feels more aesthetically pleasing. But someone might need to read the shuffled array afterwards. So I left it in.

I think it's definitely necessary to have access later. The use case of shuffled NFTs is one example as you wouldn't want to waste aux space by storing the permutation twice.

> * Prefer to derive `permutationSlot` strictly from `$.slot` instead of a user-specified `keySlot`. Else it feels dangerously easy to create two instances of `LazyShuffler` at different slots (which holds their respective `length` and the `numShuffled`), but their internal `permutation` arrays are pointing to the same slot.

Agree

> 
> I've excluded the tests to check if this works with both 16 and 32 bit numbers, cuz they take quite long to run.

src/utils/LibPRNG.sol
```diff
@@ -213,4 +240,133 @@ library LibPRNG {
...
     }
+
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*       STORAGE-BASED RANGE LAZY SHUFFLING OPERATIONS        */
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Initializes the state for lazy-shuffling the range `[0..n)`.
+    /// Reverts if `n == 0 || n >= 2**32 - 1`.
+    /// Reverts if `$` has already been initialized.
+    /// If you need to change the length after initialization, just use a fresh new `$`.
```
aschlosberg, line 251: There's conceptually a use case where the length can be extended, but never reduced (because a high value may have already been chosen). I've previously had a long-running mint during which random (already-minted) tokens were chosen for some reward. This could be implemented with a `LazyShuffler` by increasing `n` to `totalSupply()` and then calling `next()`.

Not sure it needs to be included in this PR though.

src/utils/LibPRNG.sol
```diff
@@ -213,4 +240,133 @@ library LibPRNG {
...
+            result := iszero(iszero(sload($.slot)))
+        }
+    }
+
+    /// @dev Returns if there are any more elements left to shuffle.
+    function finished(LazyShuffler storage $) internal view returns (bool result) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            let state := sload($.slot) // The packed value at `$`.
+            result := eq(shr(224, state), and(0xffffffff, state))
```
aschlosberg, line 309: Should this also check that `state != 0`?

test/LibPRNG.t.sol
```diff
@@ -233,4 +236,208 @@ contract LibPRNGTest is SoladyTest {
...
+            }
+            if (n > 32) {
+                bool anyShuffled;
+                for (uint256 i; i != n && !anyShuffled; ++i) {
+                    anyShuffled = outputs[i] != i;
+                }
+                assertTrue(anyShuffled); // Super unlikely to fail.
+            }
+            LibSort.sort(outputs);
+            for (uint256 i; i != n; ++i) {
```
aschlosberg, line 269: Out of interest, is there a reason you use `i != n` and not `i < n`?

test/LibPRNG.t.sol
```diff
@@ -233,4 +236,208 @@ contract LibPRNGTest is SoladyTest {
...
+    }
+
+    function testLazyShufflerGet() public {
+        _lazyShuffler0.initialize(16);
+        _lazyShuffler1.initialize(32);
+        uint256[] memory outputs0 = new uint256[](16);
+        uint256[] memory outputs1 = new uint256[](32);
+        unchecked {
+            for (uint256 i; i != 16; ++i) {
+                assertEq(this.lazyShuffler0Get(i), i);
```
aschlosberg, line 338: Why not just call `_lazyShuffler0.get(i)`?

src/utils/LibPRNG.sol
```diff
@@ -3,7 +3,25 @@ pragma solidity ^0.8.4;
 
 /// @notice Library for generating pseudorandom numbers.
 /// @author Solady (https://github.com/vectorized/solady/blob/main/src/utils/LibPRNG.sol)
+/// @author LazyShuffler based on NextShuffler by aschlosberg at divergencetech
```
aschlosberg, line 6: Please can you change it to this?

`/// @author LazyShuffler based on NextShuffler by aschlosberg (divergencearran)`

My GitHub pre-dates my web3 days.

src/utils/LibPRNG.sol
Vectorized: This is a legit use case. I'd like to add it.

test/LibPRNG.t.sol
Vectorized: Oh yes.

src/utils/LibPRNG.sol
Vectorized: yes

src/utils/LibPRNG.sol
```diff
@@ -213,4 +240,133 @@ library LibPRNG {
...
+    /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
+
+    /// @dev Initializes the state for lazy-shuffling the range `[0..n)`.
+    /// Reverts if `n == 0 || n >= 2**32 - 1`.
+    /// Reverts if `$` has already been initialized.
+    /// If you need to change the length after initialization, just use a fresh new `$`.
+    function initialize(LazyShuffler storage $, uint256 n) internal {
+        /// @solidity memory-safe-assembly
+        assembly {
+            if iszero(lt(sub(n, 1), 0xfffffffe)) {
```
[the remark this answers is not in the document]
Vectorized: That expression also reverts for `n == 0`.

src/utils/LibPRNG.sol
Vectorized: Added a check to revert for `finished` if `state` isn't initialized.

src/utils/LibPRNG.sol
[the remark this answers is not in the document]
aschlosberg: Nice!

PR #907, opened 2024-04-23
♻️ Use DOMAIN_SEP_B for nested EIP-712 workflow

Vectorized:
## Description

Changes:

1. Use `DOMAIN_SEP_B` (the domain separator of the app) for nested EIP-712 workflow. This means the smart account will masquerade as the app in the signature request . 
   - To make the hash un-replayable, we will add in the `eip712Domain` fields to the parent struct.
   - For phishing resistance, we will reconstruct a semi-fixed type hash on-the-fly, to enforce that some fields can't be changed to bytes32.
  
2. Make the personal sign workflow use a fixed type hash. 

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


livingrockrises:
for modular smart accounts, signature scheme is defined in the validator and gets passed on like [this](https://github.com/erc7579/erc7579-implementation/blob/main/src/MSAAdvanced.sol#L272)

in current implementation of 1271, child account require to implement below
`function _erc1271Signer() internal view virtual returns (address);`

imo account should have EIP712 base rather than validator having a EIP712 domain. how should we think about this?

Vectorized:
@livingrockrises 

Solady's implementations are not designed with modularity from the get go. This is a choice for simplicity.

This ERC1271 implementation's aim is to provide a minimal replay-safe example with production ready UX.

Once this implementation is finalized, module authors can copypaste relevant portions to create a 1271 module that implement the same behavior.

src/accounts/ERC1271.sol
```diff
@@ -190,7 +208,7 @@ abstract contract ERC1271 is EIP712 {
                 mstore(0x00, 0x1626ba7e) // `isValidSignature(bytes32,bytes)`.
                 mstore(0x20, gasBurnHash)
                 mstore(0x40, 0x40)
-                let gasToBurn := add(100000, gaslimit())
+                let gasToBurn := or(add(100000, gaslimit()), gaslimit())
```
Vectorized, line 211: I know geth has a hardcoded limit on gaslimit, but just to be double-safe.

src/accounts/ERC1271.sol
```diff
@@ -135,40 +148,45 @@ abstract contract ERC1271 is EIP712 {
         internal
         view
         virtual
-        returns (bool)
+        returns (bool result)
     {
+        (string memory name, string memory version) = _domainNameAndVersion();
```
Vectorized, line 153: It's ok to load these here regardless of whether we are using personal sign. 

We expect that most signatures won't be be using personal sign, so this is a happy path optimization.

src/accounts/ERC1271.sol
```diff
@@ -92,10 +100,13 @@ abstract contract ERC1271 is EIP712 {
...
     /// ```
-    ///     keccak256(\x19\x01 || DOMAIN_SEP_A ||
+    ///     keccak256(\x19\x01 || DOMAIN_SEP_B ||
     ///         hashStruct(Parent({
     ///             childHash: keccak256(\x19\x01 || DOMAIN_SEP_B || hashStruct(originalStruct)),
-    ///             child: hashStruct(originalStruct)
+    ///             child: hashStruct(originalStruct),
+    ///             account: address(this),
+    ///             name: keccak256(eip712Domain().name),
+    ///             version: keccak256(eip712Domain().version)
```
frangio, line 109: To make this more uniform across wallets I would make this a nested object like:
```suggestion
    ///             walletDomain: { verifyingAddress: address(this), name: ... },
```
where `walletDomain` is exactly the wallet domain as reported by its EIP-5267 `eip712Domain()` function.

Vectorized: I like this. The consistency makes it easier to standardize.

Might call it a `accountDomain` instead. As wallet might refer to a group of accounts from the same seed.

I’m imagining a public view function called `accountDomainStructHash()` that returns the bytes32 for this.

Vectorized:
Note:

Since the typehash of `Parent` is defined by the frontend, is possible for a phisher to use the final `bytes32` hash instead of passing in the values of `AccountDomain` and `Child` into the typed data signature request. 

But this is similar to asking the user to sign opaque hashes. Users should exercise precaution when seeing opaque `bytes32` hashes in the signature request. Wallets can implement heuristics to detect such cases and show a warning.

src/accounts/ERC1271.sol
```diff
@@ -130,45 +144,89 @@ abstract contract ERC1271 is EIP712 {
...
+                // Fill in the missing fields of the `TypedDataSign`.
+                mstore(t, keccak256(m, sub(add(p, c), m))) // `TYPED_DATA_SIGN_TYPEHASH`.
+                mstore(add(t, 0x20), hash) // `hash`.
+                mstore(add(t, 0x40), calldataload(add(o, 0x20))) // `contents`.
+                // The "\x19\x01" prefix is already at 0x00.
+                // `DOMAIN_SEP_B` is already at 0x20.
+                mstore(0x40, keccak256(t, 0x140)) // `hashStruct(typedDataSign)`.
+                // Compute the final hash, corrupted if contents name begins with lowercase alphabet.
+                // forgefmt: disable-next-item
+                hash := keccak256(0x1e, add(0x42, gt(0x1a, sub(byte(0, mload(add(0x1b, m))), 0x61))))
```
Vectorized, line 195: Not sure if we need the check on contents name.

Vectorized:
I'll merge this PR first, so downstream can bump their Soladys.

PR #909, opened 2024-04-24
✨ P256

z0r0z:
## Description

Begins `P256` verification path in `SignatureCheckerLib`. See: https://eips.ethereum.org/EIPS/eip-7212

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/utils/SignatureCheckerLib.sol
```diff
@@ -486,6 +487,50 @@ library SignatureCheckerLib {
...
+    /// for an ERC7212 verification in the “secp256r1” elliptic curve. Adjusts precompile for rollups.
+    /// Modified from Daimo P256 (https://github.com/daimo-eth/p256-verifier/blob/master/src/P256.sol).
+    function isValidERC7212SignatureNow(bytes32 hash, uint256 r, uint256 s, uint256 x, uint256 y)
+        internal
+        view
+        returns (bool isValid)
+    {
+        uint256 p256;
+        if (block.chainid != 1) p256 = 0x100;
+        else p256 = 0x0b;
```
Vectorized, line 504: Wait, where did you get this `0x100` from?

I heard there's a Solidity contract somewhere tho.

z0r0z: `0x100` is from the *rollup proposal* (same proposal, authors): `RIP-7212`: https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md

`0x0b` is from the mainnet version: `the next available address in the precompiled address set`: https://eips.ethereum.org/EIPS/eip-7212

So it would appear that we can't have nice things. The deployed version at `0xc2b78104907F722DABAc4C69f826a522B2754De4` could be included. It is just very much more expensive ofc as a progressive and not actual precompile. So if this library utility for P256 makes sense, I would favor going for just the precompile (being adopted by poly and arb etc.).

Vectorized: I'll put this in a `P256.sol` file instead.

Reason being contracts that need to use it will have bespoke logic for it.

Also, I don't want to deal with all the combinations of P256 + ERC1271 fallback for now.

Can you check that the gas costs of calling `0xc2b78104907F722DABAc4C69f826a522B2754De4` is indeed much more than calling `0x100`?

I'll avoid using `0xb` until the ERC is finalized.

PR #923, opened 2024-05-03
✨ Add ERC6492 unwrapping to ERC1271

Vectorized:
## Description

https://eips.ethereum.org/EIPS/eip-6492

Gonna let this marinate for a while.

**Note:** 

It's ok even if the signature is too small to decode properly.

If a certain calldata exists such that a bad decoding leads the signature verification to change from fail to pass, it will be still possible fix the calldata to make the signature verification pass in the first place (since all the required information is already in the calldata publicly).

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/accounts/ERC1271.sol
```diff
@@ -57,6 +57,20 @@ abstract contract ERC1271 is EIP712 {
...
+            if eq(
+                calldataload(add(signature.offset, sub(signature.length, 0x20))),
+                mul(0x6492, div(not(mload(0x60)), 0xffff)) // `0x6492...6492`.
+            ) {
+                let o := add(signature.offset, calldataload(signature.offset))
+                o := add(o, calldataload(add(o, 0x40)))
+                signature.length := calldataload(o)
+                signature.offset := add(o, 0x20)
+            }
+        }
```
Vectorized, line 73: It's ok even if the signature is too small to decode properly.

If a certain calldata exists such that a bad decoding leads the signature verification to change from fail to pass, it will be still possible fix the calldata to make the signature verification pass in the first place (since all the required information is already in the calldata publicly).

PR #928, opened 2024-05-09
⚡️ Optimize SafeCastLib

Vectorized:
## Description

@shuhuiluo Thanks for the inspiration.

Key ideas:

- `iszero(shr(s, add(y, x))` is cheaper than `iszero(eq(signextend(b, x), x))`.
- Using `something == uint256(0)` with early return gives cleaner bytecode for this case.

## Checklist

Ensure you completed **all of the steps** below before submitting your pull request:

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?

_Pull requests with an incomplete checklist will be thrown out._

PR #929, opened 2024-05-09
⚡️ Optimize SafeCastLib again

shuhuiluo:
## Description

Reverting some changes in #928 results in less runtime gas. No clue yet. One may need to look at the bytecode to figure out.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


Vectorized:
Very interesting. Looks like the compiler chooses to optimize differently for different sizes. Also depends on the optimization runs.

Wonder if there is a way to force the compiler to behave properly.

Vectorized:
I think this is the right idea. Different bitwidths have different optimal safe casting methods. 

Maybe `optimizer_runs` in `foundry.toml` should be set to 200, as that is the most common setting.

shuhuiluo:
> I think this is the right idea. Different bitwidths have different optimal safe casting methods.
> 
> Maybe `optimizer_runs` in `foundry.toml` should be set to 200, as that is the most common setting.

It seems we're facing optimization fragmentation. I would say people who set the optimizer runs to 200 are less concerned about a difference of 2 in runtime gas, while people like me who always use 4294967295 runs would want to minimize runtime gas, i.e. returning in the else branch for `uint256` to `uint` conversion and `uint256` to `int` conversion.

Vectorized:
Let's just use the style that is most efficient for 4294967295 runs.

PR #956, opened 2024-06-11
🥢 Chore(natspec): ERC1155 function names

chibitanaka:
With tender care, I ran `sed -i 's/Reveived/Received/g' ./src/tokens/ERC1155.sol`, each keystroke a gentle touch of affection.

In that moment, my heart swelled with a deep affection for the elegance of code, the harmony of syntax, and the beauty of precision.


Vectorized:
Incredible find.

Issue #962, opened 2024-06-17
✨ Feature request: utility function to check if a string contains only alphanumeric characters


PaulRBerg:
## Rationale

Onchain generation of NFT SVGs is on the rise. Many SVGs rely on third-party string data, e.g. ERC-20 symbols.

To sanitize strings and prevent XSS attacks, developers should only allow alphanumeric strings in the token symbol[^1]. This should be enough, since the [vast majority of tokens](https://etherscan.io/tokens) don't contain any special symbols.

It would thus be helpful to have a utility function in Solady for checking whether a string contains only alphanumeric characters.

## Example Implementation

```solidity
/// @notice Checks whether the provided string contains only alphanumeric characters and spaces.
/// @dev Note that this returns true for empty strings, but it is not a security concern.
function isAlphanumeric(string memory str) internal pure returns (bool) {
    // Convert the string to bytes to iterate over its characters.
    bytes memory b = bytes(str);

    uint256 length = b.length;
    for (uint256 i = 0; i < length; ++i) {
        bytes1 char = b[i];

        // Check if it's a space or an alphanumeric character.
        bool isSpace = char == 0x20; // space
        bool isDigit = char >= 0x30 && char <= 0x39; // 0-9
        bool isUppercase = char >= 0x41 && char <= 0x5A; // A-Z
        bool isLowercase = char >= 0x61 && char <= 0x7A; // a-z
        if (!(isSpace || isDigit || isUppercase || isLowercase)) {
            return false;
        }
    }
    return true;
}
```

[^1]: See, for example, [finding M-01](https://www.codehawks.com/report/clvb9njmy00012dqjyaavpl44#M-04) in Sablier's recent audit contest on CodeHawks.

PaulRBerg:
Alternatively, a utility function to check if a single character is alphanumeric would also be helpful:

```solidity
function isAlphanumericChar(bytes1 char) internal pure returns (bool) {
    bool isSpace = char == SPACE;
    bool isDigit = char >= ZERO && char <= NINE;
    bool isUppercaseLetter = char >= A && char <= Z;
    bool isLowercaseLetter = char >= a && char <= z;
    return isSpace || isDigit || isUppercaseLetter || isLowercaseLetter;
}
```

Vectorized:
Good feature request. Will add, thanks.

atarpara:
@PaulRBerg you request is fulfilled check it out.

0xCLARITY:
[...]

PR #963, opened 2024-06-18
✨ Add `is7BitASCII(s, allowed)` to LibString

atarpara:
## Description
closing #962 

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


src/utils/LibString.sol
```diff
@@ -406,6 +406,36 @@ library LibString {
         }
     }
 
+    /// @dev Returns if this string is a alphanumeric character including space.
```
PaulRBerg, line 409: For clarity, I would rewrite the NatSpec to:

```
/// @dev Returns true if the string contains only alphanumeric characters and spaces.
```

src/utils/LibString.sol
```diff
@@ -406,6 +406,36 @@ library LibString {
...
+                        break
+                    }
+                    o := add(o, result)
+                    if iszero(lt(o, end)) { break }
+                }
+            }
+        }
+    }
+
+    /// @dev Returns if this character is a alphanumeric including space.
```
PaulRBerg, line 431: For clarity, I would rewrite the NatSpec to:

```
/// @dev Returns true if the character is an alphanumeric character or a space.
```

PaulRBerg (approved):
very nice, thank you

0xCLARITY:
Left a more detailed comment on the issue: https://github.com/Vectorized/solady/issues/962#issuecomment-2176574758, but I think it would be _surprising_ for spaces to be considered "alphanumeric".

Vectorized:
Refactored to make it more general.

Will merge in 24 hours if no hard objections.

Vectorized (approved):

atarpara:
@Vectorized In the current implementation of the `is7BitASCII(string memory s, uint128 allowed)` function, The behavior for empty strings is to return true in all cases. This is inconsistent with the behavior observed in similar functions across various programming languages. 

For instance:

Python [isalnum](https://github.com/python/cpython/blob/main/Objects/clinic/unicodeobject.c.h#L603-L604)
Rust's [is_ascii_alphanumeric](https://doc.rust-lang.org/beta/std/primitive.char.html#method.is_ascii_alphanumeric)
Govalidator's [alphanumeric](https://github.com/asaskevich/govalidator/blob/master/patterns.go#L16)

In these implementations, the function typically returns false for an empty string. To align with these standards and ensure consistent behavior, I propose the following adjustments:

1. The function will return false for empty strings, aligning with the default behavior in other languages.
2. If allowing empty strings is necessary for specific use cases, a separate function can be created. Alternatively, developers can handle this explicitly by checking the length of the string using `bytes(str).length == 0`.

Everyone is open to give your valuable suggestions.

Vectorized:
Let's deviate away from the popular languages.

I find it easier to reason if an empty string evaluates to true.

PR #1016, opened 2024-07-16
✨ Restore counterfactual SSTORE2

emo-eth:
## Description

I'm sure you had your reasons for migrating to the CREATE3 SSTORE2 flow, but I humbly submit this PR to restore the old functionality alongside the new 😇 I have a few personal projects I plan on using counterfactual data contracts for and would love to use the latest & greatest Solady. 

Issue #998 replaced the "traditional" CREATE2-SSTORE2 flow with a "CREATE3" flow, meaning the address SSTORE2'd data will never depend on the data itself. This means that it is no longer possible to deploy "counterfactual" data with Solady, ie, data that may-or-may-not exist onchain.

Counterfactual data contracts allow you to do things like the following:
- immutably commit to a blob of data without uploading it until it is needed (e.g. onchain metadata or something akin to bitcoin taproot scripts)
- immutably commit to a blob of data with the intention of revealing its contents later (e.g. make predictions onchain without revealing their contents until later, or commit to secret onchain metadata)
- incentivize others to bear the costs of uploading a specific blob of data (e.g. tokenize and incentivize/crowdfund file chunks for something like [ethfs.xyz](https://ethfs.xyz))

Seeing as "naming things" is one of the harder problems in computer science, I'm not sure how to distinguish the "old" deterministic functions from the CREATE3 style, especially since `write` takes the same function parameters. So for now I've renamed them to be `writeCounterfactual` etc. 

I copied (and renamed) the old writeDeterministic test.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


lib/ds-test
emo-eth, line 1: uhh, not sure why this got updated, i just ran `forge build` 😕

Vectorized:
Holy sheet. This is genius.

Vectorized (approved):

src/utils/SSTORE2.sol
```diff
@@ -20,6 +20,10 @@ library SSTORE2 {
     bytes32 internal constant CREATE3_PROXY_INITCODE_HASH =
         0x21c35dbe1b344a2488cf3321d6ce542f8e9f305544ff09e4993a62319a497c1f;
 
+    /// @dev We skip the first byte as it's a STOP opcode,
+    /// which ensures the contract can't be called
+    uint256 internal constant DATA_OFFSET = 1;
+
```
Vectorized, line 26: I'll make a separate PR to remove this , as the assembly tricks reads easier if 1 is directly used in the code. Personal taste. Contextual locality.

PR #1073, opened 2024-09-13
✨ DynamicArrayLib

Vectorized:
## Description

✨

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge snapshot`?
- [x] Ran `forge test`?


atarpara (commented):
Looks Good Everything. But i think we can add `pop` method.

src/utils/DynamicArrayLib.sol
```diff
@@ -0,0 +1,453 @@
...
+            for { let arrData := mload(array) } gt(minimum, mload(arrData)) {} {
+                let w := not(0x1f)
+                // Some random prime number to multiply `cap`, so that
+                // we know that the `cap` is for a dynamic array.
+                // Selected to be larger than any memory pointer realistically.
+                let prime := 8188386068317523
+                let cap := mload(add(arrData, w)) // `mload(sub(arrData, w))`.
+                // Extract `cap`, initializing it to zero if it is not a multiple of `prime`.
+                cap := mul(div(cap, prime), iszero(mod(cap, prime)))
+                let newCap := shl(5, add(1, minimum))
```
atarpara, line 98: Why reserving `1` extra slot?

Vectorized: Cuz:

```solidity
            // Expand / Reallocate memory if required.
            // Note that we need to allocate an extra word for the length.
            for {} iszero(lt(newArrBytesLen, cap)) {} {
```

Vectorized:
If we add `pop()`, we may have to add the other variants like `popBool()`

PR #1097, opened 2024-09-30
✨ Feature: URI Component Encoding

johnnyshankman:
## Description

Implement an exact replicate of `encodeURIComponent` and `decodeURIComponent` from Javascript, as it is useful for concatenating raw string data into URIs.

Included as many tests for edge cases as I could think of.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

## Reference

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

### DIFF
--- a/src/utils/URI.sol
+++ b/src/utils/URI.sol
@@ -3,7 +3,6 @@ pragma solidity ^0.8.4;
 
 /// @notice Library to encode and decode strings in URI format.
 /// @author Solady (https://github.com/vectorized/solady/blob/main/src/utils/URIEncode.sol)
-/// @author John Shankman aka White Lights - johnny@white-lights.net (whitelights.eth)
 library URI {
     /**
      * @dev URI Encoding/Decoding Hex Table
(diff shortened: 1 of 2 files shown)


Vectorized:
Interesting work. If I may, what’s the use case?

If there’s a good use case, we’ll rewrite it in fully optimised form and add it to LibString.

On authorship, we’ll just use Solady. We don’t usually add individual authors, unless it is a link to some kind of article that describes the full algorithm, or a link to a simpler reference implementation. Otherwise, the files will be choked full with authors.

johnnyshankman:
@Vectorized sounds good!

also great question for use case.

for me? on-chain `animation_url` Data URI creation and on-chain `tokenURI` Data URI creation as well. if you want to use `data:application/json;utf8,...` all things after the utf8 must be URI encoded. the `animation_url` content must be double encoded, so having this abstracted away into a function is useful.

```
// for a tokenURI
string.concat(
    "data:application/json;utf8,",
    compiler.encodeURI(allTheJsonAsAString),
)
```

updating PR real quick to remove my author tag.

Vectorized:
Ok, let’s add this.

johnnyshankman:
@Vectorized hype! let me know if i can help at all with moving it over to libstring and optimizing. curious how you approach optimizing, good learning experience for me.

PR #1103, opened 2024-10-02
fix README.md

nnsW3:
Hello
I fixed a typo.


Vectorized:
Good find!

But let me think about it.

The milady community intentionally spells the plural as miladys instead of miladies.

This typo could be intentionally left in for the lore.

Vectorized:
Closing. Let’s keep the current spelling for the lore.

PR #1212, opened 2024-12-06
✨ STL: totalSupply

z0r0z:
Getter convenience to efficiently and safely return ERC20 supply.


Vectorized:
I'll add a revert tho. Feels safer, cuz otherwise it is not possible in differentiating between a deployed token that has no supply yet vs an empty contract.

PR #1226, opened 2024-12-13
🥢 ERC20:  _givePermit2InfiniteAllowance: `true`

z0r0z:
## Description

Describe the changes made in your pull request here.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/src/tokens/ERC20.sol
+++ b/src/tokens/ERC20.sol
@@ -665,6 +665,6 @@ abstract contract ERC20 {
     /// or else the actual allowance values may not match with the {Approval} events.
     /// For best performance, return a compile-time constant for zero-cost abstraction.
     function _givePermit2InfiniteAllowance() internal view virtual returns (bool) {
-        return false;
+        return true;
     }
 }


Vectorized:
Dang, need to update the tests too.

Vectorized:
@atarpara help me update the tests

z0r0z:
based thx @atarpara

atarpara:
Updated test, but I do not favor this because it will increase gas by default. The Solady ERC20 already has the permit functionality. So very less people use permit2.

z0r0z:
Most tokens will be used on L2 such that the convenience might be indeed worth the tradeoff. If we are shooting for hyperoptimization, then I would suggest we shuck off permit() and make it into an inheritable.

Vectorized:
Hmm… good point. @atarpara 

This extra gas is very concerning. 

Let’s marinate on this for at least a week.

Vectorized:
We can have both. For max compat.

Vectorized:
Ok, we shall only consider the gas overhead in the `transferFrom` function.

With Permit2 enabled, it costs about 20 more gas. An extra 21000 gas for the approve transaction is 1000 times more than this overhead. Hence, on overall, we save more gas. @atarpara

Vectorized:
I'll merge this when the Spearbit audit has been finalized.

PR #1237, opened 2024-12-17
♻️ Safe calldata copying

Vectorized:
## Description

Godzamm.

This is actually required, cuz calldata slicing in plain Solidity will return slices that may not have the length immediately before the data.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?


Vectorized:
@atarpara If you can think of a way to add the tests for this, lemme know.

Otherwise, its ok. The ERC1155 case is super hard to test.

PR #1243, opened 2024-12-18
✨ LibBytes cmp

Vectorized:
## Description

✨

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/src/utils/LibBytes.sol
+++ b/src/utils/LibBytes.sol
@@ -612,6 +612,35 @@ library LibBytes {
         }
     }
 
+    /// @dev Returns 0 if `a == b`, -1 if `a < b`, +1 if `a > b`.
+    /// If `a` == b[:a.length]`, and `a.length < b.length`, returns -1.
+    function cmp(bytes memory a, bytes memory b) internal pure returns (int256 result) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            let aLen := mload(a)
+            let bLen := mload(b)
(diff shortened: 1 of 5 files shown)


atarpara (approved):

src/utils/LibBytes.sol
```diff
@@ -621,19 +621,22 @@ library LibBytes {
...
                 }
             }
+            // forgefmt: disable-next-item
             if iszero(result) {
-                let x := and(not(shr(shl(3, sub(aLen, n)), not(0))), mload(add(add(a, 0x20), n)))
-                let y := and(not(shr(shl(3, sub(bLen, n)), not(0))), mload(add(add(b, 0x20), n)))
+                let x := and(mload(add(add(a, 0x20), n)),
+                    not(shr(mul(7, sub(aLen, n)), shr(sub(aLen, n), not(result)))))
+                let y := and(mload(add(add(b, 0x20), n)),
+                    not(shr(mul(7, sub(bLen, n)), shr(sub(bLen, n), not(result)))))
```
Vectorized, line 639: This is because i want this function to be able to handle cases where the string length is some crazy big amount, even tho it is like not possible unless people use inline assembly to create their strings.

Still marinating, might just add a comment about ensuring that all strings have lengths less than `2**64`.

Vectorized: The benefit of using this is that for cases where someone needs to compare a string against a possible malicious input, they can still do it without the function going into undefined behavior.

PR #1250, opened 2024-12-23
🥢 Also re-use return value in `at(AddressSet,uint256)`

0xLightt:
## Description

In EnumerableSetLib the function `at(Bytes32Set,uint256)` re-uses the return value. But `at(AddressSet,uint256)` doesn't. If there isn't any specific reason not to, both functions should do the same.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?


Vectorized:
Nice gas opt spot ;)

0xLightt:
Sorry, I ran the tests in the wrong branch... :facepalm:

PR #1306, opened 2025-01-16
✨ Add commonPrefix

JhChoy:
## Description

The `commonPrefix` function efficiently finds the common prefix (matching most significant bits) between two 256-bit numbers and returns it in hex format.

Potential Variations:
1. `commonBitPrefix` - Could return just the number of matching bits (similar to current implementation's internal use of `clz`)
2. `commonBytePrefix` - Could operate at byte granularity instead of nibbles (4 bits), which might be more useful for byte-oriented data like addresses or byte arrays

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/src/utils/LibBit.sol
+++ b/src/utils/LibBit.sol
@@ -117,6 +117,17 @@ library LibBit {
         }
     }
 
+    /// @dev Returns the common prefix of `x` and `y` in hex format.
+    function commonPrefix(uint256 x, uint256 y) internal pure returns (uint256 r) {
+        uint256 lz = clz(x ^ y);
+        assembly {
+            let nibbles := div(lz, 4)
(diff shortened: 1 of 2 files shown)


JhChoy:
Seems like gas-diff CI failed b/c the branch of my repo also `main`.
Is it better to open a new PR with another branch name?

atarpara:
@JhChoy can you provide a use case of this?

Vectorized:
Don’t worry about the CI.

Can you list some possible use cases? 

Asking cuz the audit is nearly finalized.

JhChoy:
There aren’t that many use cases. However, my initial thought was that it could be used to build radix trees or Patricia trees.

Vectorized:
@atarpara don’t close yet, lemme marinate.

PR #1315, opened 2025-01-19
✨ EnumerableKeyValueSetLib

0xLightt:
## Description

This pull request introduces an optimized version of `EnurableMapLib` designed for scenarios where both the key and value can fit into a single slot. Currently, it supports `address` keys mapped to `uint96` values.

If this functionality is deemed useful, I am open to adding more tests and functions to ensure it aligns with the existing `EnurableMapLib`. Any input or feedback would be greatly appreciated!

It is also possible to add a new entry to prep, allowing for combinations that fit within a single slot.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/src/utils/EnumerableKeyValueSetLib.sol
+++ b/src/utils/EnumerableKeyValueSetLib.sol
@@ -1,0 +1,314 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {EnumerableSetLib} from "./EnumerableSetLib.sol";
+
+library EnumerableKeyValueSetLib {
+    /*´:°•.°+.*•´.*:˚.°*.˚•´.°:°•.°•.*•´.*:˚.°*.˚•´.°:°•.°+.*•´.*:*/
+    /*                       CUSTOM ERRORS                        */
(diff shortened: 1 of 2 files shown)


Vectorized:
This is tempting.

Lemme marinate.

Vectorized:
@0xLightt Are you using this in your project or planning to use this in an upcoming project?

Vectorized:
Taste wise, I think this should be `AddressToUint96Map` in `EnumerableMapLib`.

Another problem is the audit has concluded, and `EnumerableSet` is kind of a hard file. I can activate full-effort God-eye to ensure that there's no bug, but you know... public perception. Not everyone trusts my spidersense.

0xLightt:
I am not using exactly this. It's slightly modified because we only have one instance per contract and don't have any zero addresses or values. So we can skip the whole zero sentinel logic and also just have 3 functions: 'values' and both 'addOrUpdate' and 'remove' functions that return the previous values. I will gladly share it, but it's too tailored to add to solady.

Appreciate the naming suggestion! ❤️

It goes without saying, but it's totally up to you. In my opinion, if there is to be an address to uint96 enumerable map it will eventually be optimized to use 1 slot. If the risk or effort isn't worth optimizing it at the moment, then I think that what already exists is more than enough.

4player90 (approved):

PR #1328, opened 2025-01-28
📝 Updated Copyright Year

maximevtush:
Updated the copyright year from 2022 to 2025 in the LICENSE file

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2025 Solady.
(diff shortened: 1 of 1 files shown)


LICENSE.txt
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2022 Solady.
+Copyright (c) 2025 Solady.
```
atarpara, line 3: It should be `2022-2025`. 

@Vectorized what's your thought?

Vectorized:
Ah ok. Let’s update.

PR #1245, opened 2024-12-19
♻️ Make ERC721 overflow check safer

Vectorized:
## Description

♻️

## Checklist

Ensure you completed **all of the steps** below before submitting your pull request:

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

_Pull requests with an incomplete checklist will be thrown out._

PR #1335, opened 2025-02-01
♻️ Revert to the Original Implementation of ERC721

atarpara:
## Description

Reverts #1245 changes

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/src/tokens/ERC721.sol
+++ b/src/tokens/ERC721.sol
@@ -294,14 +294,14 @@ abstract contract ERC721 {
             {
                 mstore(0x00, to)
                 let toBalanceSlot := keccak256(0x0c, 0x1c)
-                let p := sload(toBalanceSlot) // `toBalanceSlotPacked`.
-                // Revert if `to` is the zero address, or if the account balance is maxed.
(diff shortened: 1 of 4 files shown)


atarpara:
@kadenzipfel can you please again review it.

Vectorized (approved):

Vectorized:
@atarpara also update the zksync variant hehe

Vectorized:
In halmos we trust.

kadenzipfel:
Why is this being reverted?

kadenzipfel:
I assume this is being reverted due to the low likelihood of the issue occurring in practice? If so, I don't see any other issue with this logic, but I would recommend at least adding some documentation to explain the circumstance where the overflow check can fail

Fwiw I think it's worth it to have the extra safe overflow check, especially since I think it only saves about 3 gas

Vectorized:
@kadenzipfel The overflow check will never fail. 

`type(uint256).max + 1 == 0`.

kadenzipfel:
> @kadenzipfel The overflow check will never fail.
> 
> `type(uint256).max + 1 == 0`.

Big oof lol. Guess we're all good then!

PR #1353, opened 2025-02-11
♻️ More memory-safe annotations

Vectorized:
## Description

https://docs.soliditylang.org/en/latest/assembly.html#advanced-safe-use-of-memory

Beyond the strict definition of memory-safety given above, there are cases in which you may want to use more than 64 bytes of scratch space starting at memory offset 0. If you are careful, it can be admissible to use memory up to (and not including) offset 0x80 and still safely declare the assembly block as memory-safe. This is admissible under either of the following conditions:

- By the end of the assembly block, the free memory pointer at offset 0x40 is restored to a sane value (i.e. it is either restored to its original value or an increment of it due to a manual memory allocation), and the memory word at offset 0x60 is restored to a value of zero.

- The assembly block terminates, i.e. execution can never return to high-level Solidity code. This is the case, for example, if your assembly block unconditionally ends in calling the revert opcode.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/src/accounts/EIP7702Proxy.sol
+++ b/src/accounts/EIP7702Proxy.sol
@@ -38,7 +38,8 @@ contract EIP7702Proxy {
     /*.•°:°.´+˚.*°.˚:*.´•*.+°.•°:´*.´•*.•°.•°:°.´:•˚°.*°.˚:*.´+°.•*/
 
     constructor(address initialImplementation, address initialAdmin) payable {
-        assembly ("memory-safe") {
+        /// @solidity memory-safe-assembly
+        assembly {
             sstore(_ERC1967_IMPLEMENTATION_SLOT, shr(96, shl(96, initialImplementation)))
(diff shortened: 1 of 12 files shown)


src/accounts/EIP7702Proxy.sol
```diff
@@ -50,6 +51,7 @@ contract EIP7702Proxy {
 
     fallback() external payable virtual {
         uint256 s = __self;
+        /// @solidity memory-safe-assembly
         assembly {
             mstore(0x40, returndatasize()) // Optimization trick to change `6040608052` into `3d604052`.
```
shuhuiluo, line 56: Do you mean `6080604052` (`mstore(0x40, 0x80)`)?

Vectorized: Oh yes. Nice find!

PR #1355, opened 2025-02-17
✨ Improve EIP7702Proxy and add LibEIP7702

Vectorized:
## Description

LibEIP7702 makes it neater to work with the proxy.

Principles:

- Try to preserve the upper 96 bits. Just in case some weird delegation saved some values in there.
- Use transient storage to pass the request. Technically, we can just use regular storage, without much overhead, cuz it's warm. But I am worried about the super edge case where someone actually use the upper 96 bits. Also it's cheaper for the case of direct delegation (no via EIP7702Proxy).

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ accounts
 ├─ ERC6551 — "Simple ERC6551 account implementation"
 ├─ ERC6551Proxy — "Relay proxy for upgradeable ERC6551 accounts"
 ├─ ERC7821 — "Minimal batch executor mixin"
+├─ LibEIP7702 — "Library for EIP7702 operations"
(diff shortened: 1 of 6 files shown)


Vectorized:
Should we add some kind of identifier to the proxy?

Vectorized:
Marinating for the next 6 hours, then merging. @atarpara

atarpara (approved):
LGTM.

atarpara:
> Should we add some kind of identifier to the proxy?

We can use proxy `codehash` as a identifier.

PR #1367, opened 2025-03-06
🥢 Update Permit2 comment

hexshire:
## Description

Updated documentation to clarify that Permit2 is enabled by default and only requires overriding to disable.
https://github.com/Vectorized/solady/blob/09cdd7c8d4b4d83ab7b0c1e313df05310c4fc920/src/tokens/ERC20.sol#L668


Vectorized:
Woah, nice find.

PR #1371, opened 2025-03-09
✨ Add `set` and `delete_` to native bytes/string in storage

shuhuiluo:
## Description

Add `set` and `delete_` functions to simplify handling of bytes and string storage references in `LibBytes` and `LibString`. These changes enable efficient and memory-safe operations when assigning or deleting storage references directly.

## Rationale

As the following example shows, Solidity only allows[ assignments to state variables or to members of local variables of storage struct type](https://docs.soliditylang.org/en/latest/types.html#data-location-and-assignment-behavior). One cannot assign a bytes array in memory/calldata to a local bytes array storage reference. More often than not, people first read from a member of an array/mapping state variable, `dataArray[i]`, and then write to the same member using `dataArray[i] = x`, which involves two slot calculations and therefore two `keccak256`s.

A workaround is to cast the bytes storage reference to a wrapper struct, which isn't obvious and clean. Besides, the bytecode for bytes assignment and deletion generated by solc is suboptimal.

While Solady already has `BytesStorage` and `StringStorage`, switching to a different storage layout than that of native bytes/string isn't always possible, which necessitates the current implementation.

```solidity
    bytes data;
    bytes[] dataArray;

    struct BytesWrapper {
        bytes inner;
    }

    function f() public {
        data = "Asterix";
        delete data;
        dataArray.push("Milady");
        dataArray[0] = "Asterix";
        delete dataArray[0];
        bytes storage y = dataArray[0];
        // Error (7407): Type literal_string "Asterix" is not implicitly convertible to expected type bytes storage pointer.
        // y = "Asterix";
        // Error (9767): Built-in unary operator delete cannot be applied to type bytes storage pointer.
        // "delete y" is not valid, as assignments to local variables referencing storage objects 
        // can only be made from existing storage objects.
        // https://docs.soliditylang.org/en/latest/types.html#delete
        // delete y;
        BytesWrapper storage bw;
        assembly {
            bw.slot := y.slot
        }
        bw.inner = "Asterix";
        delete bw.inner;
    }
```

> 

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/src/utils/LibBytes.sol
+++ b/src/utils/LibBytes.sol
@@ -33,19 +33,32 @@ library LibBytes {
         assembly {
             let len := mload(s)
             let packed := shl(1, len)
-            for { let i := 0 } 1 {} {
-                let o := add(s, 0x20)
+            for {} 1 {} {
                 if iszero(gt(len, 0x1f)) {
-                    packed := or(mload(o), packed)
+                    let right_aligned := mload(add(s, len))
+                    let zeros := shl(3, sub(0x20, len))
+                    let left_aligned := shl(zeros, right_aligned)
+                    packed := or(packed, left_aligned)
                     break
                 }
                 packed := or(packed, 1)
                 mstore(0x00, $.slot)
-                for { let p := keccak256(0x00, 0x20) } 1 {} {
-                    sstore(add(p, shr(5, i)), mload(add(o, i)))
-                    i := add(i, 0x20)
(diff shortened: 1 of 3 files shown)


Vectorized:
O.O

Vectorized:
Lemme marinate first.

May or may not merge.

PR #1425, opened 2025-06-05
✨ LibBytes indexOfByte

Hyodar:
## Description

Adds `function indexOfByte(bytes memory subject, bytes1 needle, uint256 from)` and offset-less shortcut to `LibBytes` and `LibString`.

This is a considerably more optimal version of `indexOf` in the case only a byte needle is used as we can apply some learnings from [bithacks](https://graphics.stanford.edu/~seander/bithacks.html) and keep the gas cost per 'missed' chunk quite low.

The reason for naming it `indexOfByte` versus keeping it as `indexOf` is because Solidity also tries to cast literal strings to `bytes1`, so if there's an implementation for `bytes` and one for `bytes1`, the user always has to be explicit on the type, which can get annoying.

Reproduced the `clz` implementation inside of it to avoid creating a dependency just for that and for a very small optimization, but it would be okay to do otherwise as well.

(Keeping it as draft for a review soon, and possibly some more tests.)

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?


Vectorized:
looks good. I might cache the mask in a variable to make the bytecode smaller.

PR #1462, opened 2025-08-11
📝 Correct erc20.md links in weth.md

VolodymyrBg:
Remove incorrect "tokens/" prefix from relative links since both files 
are in the same directory.


Vectorized:
Nice catch. I think we need to update the docs generator JS to somehow resolve relative links. Will do it later.

PR #1464, opened 2025-08-12
⚡️ Optimize Pow2 Padding

0xClandestine:
## Description

⚡️ Replace iterative `roundUpPow2` calculation with a constant-time bit-twiddling version inspired by [“Bit Hacks” – Round up to next power of 2](https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2).

`O(log(n)) -> O(1)`

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/src/utils/MerkleTreeLib.sol
+++ b/src/utils/MerkleTreeLib.sol
@@ -254,6 +254,7 @@ library MerkleTreeLib {
         pure
         returns (bytes32[] memory result)
     {
+        uint256 p = roundUpPow2(leaves.length);
         /// @solidity memory-safe-assembly
         assembly {
             result := mload(0x40)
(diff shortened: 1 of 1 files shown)


Vectorized:
Does the bytecode size go up or down?

0xClandestine:
> Does the bytecode size go up or down?

@Vectorized `MerkleTreeLibTest` size increased by 12 bytes after my last commit. Also thinking about it further, there's no need to fully support uint256 values with this approximation, 32 bits would sufficient to support all possible `leaves.length` values since memory has practical size limits.

src/utils/MerkleTreeLib.sol
```diff
@@ -262,8 +262,9 @@ library MerkleTreeLib {
                 mstore(0x00, 0xe7171dc4) // `MerkleTreeLeavesEmpty()`.
                 revert(0x1c, 0x04)
             }
-            let p := 1 // Padded length.
-            for {} lt(p, l) {} { p := add(p, p) }
+            let p := sub(mload(leaves), 1)
+            for { let i := 1 } lt(i, 0x80) { i := add(i, i) } { p := or(p, shr(i, p)) }
```
0xClandestine, line 266: Supporting lengths up to `type(uint32).max` should be sufficient, will let you make that choice though.
```suggestion
            for { let i := 1 } lt(i, 0x10) { i := add(i, i) } { p := or(p, shr(i, p)) }
```

Vectorized:
I can see why this might be needed onchain.

0xClandestine:
> I can see why this might be needed onchain.

I know `merkleizeSha256` is needed for beaconchain related proofs and is actively used by EigenLayer.

PR #1476, opened 2025-08-21
fix: correct relative paths in token documentation links

GarmashAlex:
Fix relative paths in token docs: change 'tokens/erc20.md' to 'erc20.md'


Vectorized:
This fix is not needed. 
- Docs are autogenerated.
- All links are absolute to `src`.

PR #1485, opened 2025-09-03
✨ Add new calldata optimal execution mode for ERC7821

legion2002:
## Description

The PR adds a new calldata optimized execution mode to ERC7821 which allows you to batch calls to the same target address, without sending the redundant "to" variable in each call.
It also removes the value field, and assumes it will be 0 for all the calls. This is usually true for most DeFi batches, because everything uses WETH.

Even if the `address(0)` optimization is used in calldata, for each call there is still a 4 gas overhead. 
With this PR, you can compress the calldata down to the theoretical optimal, this is specially useful when you want to do highly optimized self calls.

Note: The `address(0)` replacement is still available with the common `to` field passed.

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?

### DIFF
--- a/src/accounts/ERC7821.sol
+++ b/src/accounts/ERC7821.sol
@@ -264,6 +264,9 @@ contract ERC7821 is Receiver {
     function _execute(CallSansTo[] calldata calls, address to, bytes32 keyHash) internal virtual {
         unchecked {
             uint256 i;
+            if (to == address(0)) {
+                to = address(this);
+            }
             if (calls.length == uint256(0)) return;
             do {
                 (uint256 value, bytes calldata data) = _get(calls, i);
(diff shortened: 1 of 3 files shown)


Vectorized:
@howydev @legion2002 @atarpara 

Upon further thinking, this is actually more non-trivial than expected. 

First, `ERC7821` does not make any bounds checks. This is intentional (Solady's tries to cater to max performance under expert use, freedom to bear footguns).

We expect that users who require checks use `LibERC7579.decodeBatchAndOpData` and `LibERC7579.decodeBatch`. 

The reason why `ERC7821` does not have the checks is because we expect that the `executionData` will be checked once via `LibERC7579` somewhere else, when it matters. For Ithaca's case, that check is in the Orchestrator, since there's validation AND execution. If there's only execution, we don't need that check.

If we want to put this on `ERC7821`, we will need `LibERC7579.isCompactBatch` and `LibERC7579.decodeCompactBatch`.

Maybe can do this just on the Orchestrator, if all the intents will be done via the Orchestrator. I'm currently thinking if this is actually worth the effort.

PR #1522, opened 2026-05-08
implement ERC-8255

moodysalem:
## Description

Implements [ERC-8255](https://eips.ethereum.org/EIPS/eip-8255) in the ERC20 token

## Checklist

- [x] Ran `forge fmt`?
- [x] Ran `forge test`?


atarpara:
@moodysalem I don't think modifying the current `ERC20.sol` is a good approach for ERC-8255.

moodysalem:
> @moodysalem I don't think modifying the current `ERC20.sol` is a good approach for ERC-8255. 

Could you please elaborate?

There is a section in the EIP now regarding security considerations and the backwards compatibility section is more fleshed out. Together I think those sections explain why this should be the default

